PR #68, opened 2023-10-27
✨ Replace loop icon png with custom SF Symbol

caocanx:
Make the icon looks smoother. It created base on the `circle.circle` in SF Symbols 5.

Slide 16_9 - 1


MrKai77:
Looks good! Just tested this locally and the icon in the settings window looks so much smoother!
My only piece of feedback is that the menubar icon looks a little "heavy". Would you be able to make it look lighter in the menubar? (It looks perfect in the settings window)
Adding `.fontWeight(.light)` should be sufficient :)
Once that is done, this PR should be ready for merge!

Loop/MenuBar/MenuBarIconView.swift
```diff
@@ -10,7 +10,10 @@ import SwiftUI
 struct MenuBarIconView: View {
     @State var rotationAngle: Double = 0.0
     var body: some View {
-        Image(.menubarIcon)
+        Image("loop")
```
mrkai77, line 13: Add `.fontWeight(.light)`

caocanx: Very good advice, I added it.

MrKai77 (approved):
Looks good!

Issue #75, opened 2023-10-28
✨ Custom keybindings


Zabriskije:
### Feature Request Description

Although I found the current mapping pretty good, it would be nice to give the users the option to choose their custom keybindings.

This also helps with [Trigger Key conflicts](https://github.com/MrKai77/Loop/issues/74).

### Final Checks

- [X] My issue title is descriptive
- [X] This is a single feature request (multiple feature requests should be reported individually)

MrKai77:
I've been wanting to implement this for a while now too! The only problem is that I am having trouble figuring out how to configuration UI would look like. However, since Loop's UI to show the user the current keybinds isn't that good to begin with, I guess this is a good opportunity to improve it!

Zabriskije:
I think you did a good job with Loop's UI. The current Keybindings section it's spot-on for its purpose.
To give you an idea of how to implement it, here's an example from Moom's menu:

![example](https://github.com/MrKai77/Loop/assets/101254295/ba8e8906-e2f3-405d-9e17-55574aeaa789)

I don't know how complicated is the cells/tiles selection and all the other bells and whistles, but it looks hard. A simpler solution could be to just have a drop-down menu for action selection, a blank input field for keybindings, and a +/- button for adding/removing them.

MrKai77:
This does look pretty cool! I'll see what I can do!

Zabriskije:
Not sure if this needs a new issue since you're going to redo the Keybindings section, but in `beta.11` I can't use <kbd>Shift</kbd> as a trigger key. Works fine on `beta.10`.

MrKai77:
Yup, this changed in #40, where you can press <kbd>Trigger Key</kbd> + <kbd>Shift</kbd> + <kbd>Space</kbd> to toggle fullscreen mode on a window. However, it could make sense to set this back to <kbd>Trigger Key</kbd> +  <kbd>F</kbd>, so that <kbd>shift</kbd> is useable as trigger key again (and this would also simplify the new keybind system a lot :P). 

*But*, <kbd>Trigger Key</kbd> + <kbd>Shift</kbd> + <kbd>Space</kbd> also makes sense as a keybind to toggle fullscreen because <kbd>Trigger Key</kbd> + <kbd>Space</kbd> is already a keybind to maximize a window. 

Another option we could consider is being able to double-click a key (if it's a single key) as well. For example, a single click on <kbd>Space</kbd> would be maximize, but a double click could be toggle fullscreen.

[...]

PR #86, opened 2023-11-19
💄 Make the shortcut view resemble a real keyboard more closely

caocanx:
CleanShot 2023-11-19 at 16 47 59@2x


MrKai77:
I really appreciate the effort you've put into making it look so clean!
Unfortunately, #75, which is nearly complete, involves removing this view in favor of user-customizable keybinds, so I won't be able to merge your PR.

However, I want to thank you for your hard work on this. In the future, it might be helpful to open an issue first to discuss potential changes and ensure they align with ongoing developments. This way, we can coordinate more effectively :)

PR #174, opened 2024-01-16
✨ #171 Less intrusive prompt when new icon is unlocked

### LINKED ISSUE #171: ✨ Less intrusive prompt when new icon is unlocked

Currently, when the user unlocks a new icon, it shows an alert which can possibly be intrusive. Maybe there's other ways I can more quietly notify the user about new icons so that they don't get too distracted? Or maybe a toggle to notify you about new icons so that you can optionally disable them?

### Screenshots

_No response_

### Additional Context

_No response_

### Final Checks

MrKai77:


MrKai77:
What the new system looks like:
Screen Recording 2024-01-15 at 6 12 04 PM
These notifications can also be disabled as well.

JaceThings:
lol, i can't turn this on

Version 1.0.0-beta.13 (635)

https://github.com/MrKai77/Loop/assets/60551230/b7213609-a0f7-4263-a25d-06b010e60c93

MrKai77:
Whattt can you send a screenshot of your settings for notifications for Loop? 

Here's how it looks for me:
image

Looks like I will need to fix this haha

JaceThings:
Makes sense. Loop never asked me to allow it to send notifications.

Once I allowed it manually, it all works.

MrKai77:
Oh that's really weird.. I'll have to look into this!

MrKai77:
Alright, so apps are only allowed to request notification permissions if they're not already in the Notifications section in the System Settings app (probably to prevent spamming).

In your case, Loop was already somehow there even though it never asked to send notifications, which made it impossible for it to send you any further requests to send notifications. 

So what I did was make it tell you to manually enable in that case (since there's nothing much I can do other than that):
Screenshot 2024-01-21 at 4 33 13 PM

And here's the build:
[Loop.zip](https://github.com/MrKai77/Loop/files/14003195/Loop.zip)

Thanks for noticing this! This would have been a super confusing if I released it as-is :)

PR #202, opened 2024-02-04
✨ #94 Custom window padding on each side of screen

### LINKED ISSUE #94: ✨ Custom window padding on each side of screen

Loop is lovely so far. Really love the intuitive feeling of the gestures as well as the animations, I use yabai for automated window management and have some custom window size settings since I'm using apps like Ubersicht for custom task bars.

Would be great if when using Loop, they would respect these window gaps and borders.

### Screenshots

What window looks like with yabai window gaps normally:

image

What window looks like after loop resizing:

image

### Additional Context

_No response_

### Final Checks

MrKai77:
Preview: image


y-u-s-u-f:
Cool!

Issue #138, opened 2023-12-27
✨ Trackpad gestures to trigger Loop


ankidoc:
### Feature Request Description

So since I was running into a problem with the function keys cmd, option and control, I was wondering if there will be implementation of normal shortcut at any point?

The problem at my place is that even with the box ticked for the option  that I have to push town twice the key, the loop gets triggered when I have combinations like cmd + shift + space when cmd is the key that I assigned to trigger loop when pushed 2 times. I have a few shortcuts that involve the functions keys and almost every key has some 2 or 3 keybinding strokes. Why is loop getting triggered?

The reason for the above is, that I use better touch tools and I had the cool idea in mind that I can combine them.
So I make a one finger force touchpad move and assign the same shortcut that triggers loop.
Unfortunately at the moment there is no possiblity to only trigger a function key while pressing down on the track pad with force. So I was hoping that would solve the issue.

### Final Checks

- [X] My issue title is descriptive
- [X] This is a single feature request (multiple feature requests should be reported individually)

MrKai77:
Yep, the double clicking is a known bug in the latest release, and you can download a fixed version here: #127 (but I believe this build breaks all of Loop's previous keybindings).

This build should also have a "Hide until direction is chosen" toggle under the Radial Menu section of the settings window, which should help you combine this with BetterTouchTool.

Otherwise, just to make sure I understand correctly, you are hoping that a force-click on a trackpad could be used as a trigger for Loop?

ankidoc:
> Yep, the double clicking is a known bug in the latest release, and you can download a fixed version here: #127 (but I believe this build breaks all of Loop's previous keybindings).
> 
> This build should also have a "Hide until direction is chosen" toggle under the Radial Menu section of the settings window, which should help you combine this with BetterTouchTool.
> 
> Otherwise, just to make sure I understand correctly, you are hoping that a force-click on a trackpad could be used as a trigger for Loop?

exactly my point.

So just to understand (again, I'm not a coder and familiar with how git things work): There will be not one but more builds to serve different needs, or are these testbuilds and you just keep them up? because the feature of hiding until toggeling is awesome too. Commander and Charmstone are having this feature too. Why is it not in the newest version? :)

MrKai77:
These are just test builds (you can also think of them as alpha builds, which are potentially unstable) :)
I just have these builds to make sure a feature works well (and is bug-free), before fully releasing it to the public.
[...]

PR #214, opened 2024-02-14
✨ #138 Mouse gestures to trigger Loop

MrKai77:


MrKai77:
I'll be closing this PR since it is old, and many new code changes have been introduced ever since. While I can merge back from develop, there are many conflicts, and restarting the implementation of this feature would be easier :)

PR #253, opened 2024-03-13
✨ #233 Disable looping in certain apps

### LINKED ISSUE #233: ✨ Disable looping in certain apps

So i don't really want arc to be looping and i think there must be some apps that everyone have which they don't wanna snap, so it would be very nice to have a feature to disable snapping for certain apps

### Screenshots

_No response_

### Additional Context

_No response_

### Final Checks

just-doit:
Allows an exclusion list to be configured. Applications on this list are ignored when looping.
The bundle IDs of the applications are saved in an array in the UserDefaults for this purpose.

### DIFF
--- a/Loop/Managers/AppListManager.swift
+++ b/Loop/Managers/AppListManager.swift
@@ -1,0 +1,59 @@
+//
+//  AppListManager.swift
+//  Loop
+//
+//  Created by Dirk Mika on 11.03.24.
+//
+
+import SwiftUI
+import Algorithms
+
+
+class AppListManager: ObservableObject {
+    
+    struct App: Identifiable
+    {
+        var id: String { bundleID }
+        var bundleID: String
+        var displayName: String
+        var installationFolder: String
+    }
+    
+    private var qry = NSMetadataQuery()
+    
+    @Published var installedApps = [App]()
+    
+    init() {
+        self.startQuery()
+    }
+    
+    deinit {
+        qry.stop()
+    }
+    
+    private func startQuery() {
+        qry.predicate = NSPredicate(format: "kMDItemContentType == 'com.apple.application-bundle'")
+        if let appFolder = FileManager.default.urls(for: .applicationDirectory, in: .localDomainMask).first {
+            qry.searchScopes = [appFolder]
+        }
+
+        NotificationCenter.default.addObserver(forName: .NSMetadataQueryDidFinishGathering, object: nil, queue: nil, using: queryDidFinishGathering)
+        qry.start()
+    }
+    
+    private func queryDidFinishGathering(notification: Notification)
+    {
+        if let items  = qry.results as? [NSMetadataItem] {
+            self.installedApps = items.compactMap({ item in
+                guard let bundleId = item.value(forAttribute: NSMetadataItemCFBundleIdentifierKey) as? String,
+                      let displayName = item.value(forAttribute: NSMetadataItemDisplayNameKey) as? String,
+                      let path = item.value(forAttribute: NSMetadataItemPathKey) as? String,
+                      let installationFolder = URL(string: path)?.deletingLastPathComponent().absoluteString.removingPercentEncoding
+                else  {
+                    return nil
+                }
+                return App(bundleID: bundleId, displayName: displayName, installationFolder: installationFolder)
+            })
+        }
+    }
+}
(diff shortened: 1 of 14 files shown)


Loop/Window Management/WindowEngine.swift
```diff
@@ -125,6 +125,7 @@ struct WindowEngine {
     static var frontmostWindow: Window? {
         guard
             let app = NSWorkspace.shared.runningApplications.first(where: { $0.isActive }),
+            !Defaults[.applicationExcludeList].contains(app.bundleIdentifier ?? ""),
```
mrkai77: This will only work when the user has "Resize window under cursor" disabled.

Loop always will get the target window using `getTargetWindow()` (with an exception for window snapping), which when will retrieve this variable if the user has that setting disabled. However, since `getTargetWindow()` doesn't have an instance of `NSRunningApplication`, that might be pretty complicated. 

What I would recommend is to instead add a variable inside `Window`, called `isAppExcluded` (or similar), which checks if its own `nsRunningApplication`'s bundle ID is inside the excluded apps list :)

Then, inside LoopManager, you would be able to do: 
`guard self.targetWindow.isAppExcluded != true else { return }`

(let me know if you need more details, as my explanation might have been a little confusing!)

Loop/Settings/SettingsView.swift
```diff
@@ -47,8 +48,18 @@ struct SettingsView: View {
                 .frame(width: 500)
                 .frame(minHeight: 500, maxHeight: 680)
 
-            MoreSettingsView()
+            ExcludeListSettingsView()
                 .tag(5)
+                .tabItem {
+                    Image(systemName: "xmark.rectangle")
+                    Text("Black list")
```
mrkai77: Replace "Black list" with something like "Excluded Apps" so that it matches the code :)

Loop/Managers/AppListManager.swift
```diff
@@ -0,0 +1,58 @@
...
+//
+//  Created by Dirk Mika on 11.03.24.
+//
+
+import SwiftUI
+import Algorithms
+
+class AppListManager: ObservableObject {
+
+    struct App: Identifiable {
```
mrkai77, line 13: It might also be interesting to store the app's URL, so that inside the settings window, you would be able to preview the app's icon (rather than just being text)!

Look at `WallpaperView` if you need help loading an image from a file URL :)

Loop/Settings/SettingsView.swift
```diff
@@ -47,8 +48,18 @@ struct SettingsView: View {
...
+                }
+                .environmentObject(appListManager)
+                .frame(width: 500)
+                .frame(minHeight: 500, maxHeight: 680)
```
mrkai77: Unlike the keybindings tab, this tab only contains a list, and no toggle buttons etc.
So I think this might look better without a minimum height (but keep the max height)!

Loop/Settings/ExcludeListSettingsView.swift
```diff
@@ -0,0 +1,133 @@
...
+                                .buttonStyle(.plain)
+                                .padding(5)
+                            }
+                    }
+                    .ignoresSafeArea()
+                    .padding(-10)
+                } header: {
+                    VStack(alignment: .leading) {
+                        Text("Excluded applications")
+                        Text("Applications on the exclude list are ignored when looping.")
```
mrkai77: Simplify to "Applications in the exclude list are ignored by Loop."

Loop/Settings/SettingsView.swift
```diff
@@ -47,8 +48,18 @@ struct SettingsView: View {
                 .frame(width: 500)
                 .frame(minHeight: 500, maxHeight: 680)
 
-            MoreSettingsView()
+            ExcludeListSettingsView()
                 .tag(5)
+                .tabItem {
+                    Image(systemName: "xmark.rectangle")
```
mrkai77: I think the `xmark.app` SF symbol might look cleaner

MrKai77 (changes requested):
Your code is very nicely-made, love it! Here's a few changes that I think would improve this functionality :D

Loop/Managers/AppListManager.swift
just-doit: I've added the app icon as NSImage to the App struct.

MrKai77 (approved):
I made some very minor changes with the code style, but everything else looks great :)
Will merge!

Issue #278, opened 2024-03-28
🐞 Make Loop work on display's edges


phlpsong:
### Bug Description

When the  mouse is on the edge of display, the panel could not fully shown

### Steps To Reproduce

1. move the mouse to the edge of display
2. press shortcut key to trigger the panel to show
3. check panel

### Expected Behavior

show the full panel view

### Actual Behavior

only show part of the panel

### Screenshots

<img width="2560" alt="SCR-20240328-uiwl" src="https://github.com/MrKai77/Loop/assets/103433299/07767e06-ab25-4ba2-85b7-52f1718b29b0">

### MacOS Version

Sonoma 14.4.1

### Loop Version

beta.14(773)

### Final Checks

- [X] My issue title is descriptive
- [X] This is a single bug (multiple bugs should be reported individually)

MrKai77:
This is currently intended behavior. However, to make Loop work on the sides of the screen, I think that obtaining the raw mouse input from NSEvent would be beneficial, so that the user can still access the obscured radial menu items :)

SenpaiHunters:
Hey, @phlpsong. Kai and I have had a long talk over the months regarding this issue: how do we support it and make it smooth as part of Loop? We've come to a decision that, overall, we will not change or try to move the user's cursor and Loop's radial menu in order to make room. This will feel off and will require a lot of extra logic and checking, adding more overhead and issues if something goes wrong. For now, we do not plan on making any advances; however, we will look at alternative ways as we progress in Loop's development, if there are any. Thank you for your patience.

phlpsong:
Ok, it make sense. Thank for the update.

SenpaiHunters:
Okay, @phlpsong, small update. We found a way! It does not move anything in Loop, not the cursor or the preview; however, simply moving up and down the restricted section will allow the radial menus set binds to work as they would with full width! Hopefully, this helps. Thanks.

PR #279, opened 2024-03-28
Radial Menu issue when mouse on the edge of display

phlpsong:
https://github.com/MrKai77/Loop/issues/278

https://github.com/MrKai77/Loop/assets/103433299/1e9e5158-9b81-4802-9580-cafa8ef24656

### DIFF
--- a/Loop/Radial Menu/RadialMenuController.swift
+++ b/Loop/Radial Menu/RadialMenuController.swift
@@ -18,11 +18,6 @@ class RadialMenuController {
             return
         }
 
-        let mouseX: CGFloat = position.x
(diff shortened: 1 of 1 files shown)


MrKai77:
Unfortunately, even though this visually fixes it, the `initialMousePosition` is still set to the side/corner of the screen, meaning that it functionally stays the same. Rather than moving the radial menu, I would recommend capturing the raw mouse movement and using that over the cursor movement so that Loop can still be used at the sides/corners of the screen. (without moving the radial menu)

phlpsong:
OK, thanks for your comment.

PR #290, opened 2024-03-29
Add SettingTab enum type

phlpsong:
Define an enum `SettingTab` to replace the current hardcode tab index


Loop/Settings/SettingsView.swift
```diff
@@ -69,4 +69,13 @@ struct SettingsView: View {
         }
         .fixedSize(horizontal: true, vertical: true)
     }
+
+    enum SettingTab: Int {
```
mrkai77: Really small request: can you change `SettingTab` to `SettingsTab` since it's a single tab *within* multiple settings?

MrKai77 (changes requested):

Loop/Settings/SettingsView.swift
```diff
@@ -49,7 +49,7 @@ struct SettingsView: View {
                 .frame(minHeight: 500, maxHeight: 680)
 
             ExcludeListSettingsView()
-                .tag(5)
+                .tag(SettingTab.excludeList)
```
mrkai77: `excludeList` -> `excludedApps`

Loop/Settings/SettingsView.swift
phlpsong: Sure, renamed to `SettingsTab`.

MrKai77 (approved):
Thanks! Looks good!

PR #291, opened 2024-03-29
StageManager optimization

phlpsong:
It appears that the `available` property is no longer necessary, so I have removed it. 

Additionally, I have simplified the logic for obtaining other values.


MrKai77 (approved):
Looks good!

PR #391, opened 2024-06-13
✨ Use `CGEventMonitor` for trigger key detection

MrKai77:


MrKai77:
Will close since this ends up completely breaking flag-based trigger key detection 😓

PR #498, opened 2024-07-09
✨ Ability to ignore fullscreen windows

MrKai77:


SenpaiHunters (approved):
This works perfectly with full-screen apps, both games and videos, should be good to push to developmental branch.

PR #499, opened 2024-07-09
🌐 Improve localization process

MrKai77:


SenpaiHunters (approved):
Lots of changes for localisation, and more will need to be made support future buttons,, and more, but for now, this is good, everything seems to compile without any issues.

PR #500, opened 2024-07-09
✨ Import from Rectangle (+ Pro)

### LINKED ISSUE #97: ✨ Keybind presets

It would be helpful to have default key bindings (or existing presets)! So, if you're used to Rectangle's default key bindings, you should be able to toggle them from either a dropdown menu or some sort of button, and it will automatically set all key bindings to Rectangle's defaults.

Basically, JSON importing of Rectangle into Loop is another way, so the key bindings don't need to be manually set again.

### Screenshots
Rectangle Pro - November 11 2023 - 10 31 AM@2x

### Additional Context

_No response_

### Final Checks

-  [X] My issue title is descriptive
-  [X] This is a single feature request (multiple feature requests should be reported individually)

SenpaiHunters:


Loop/Luminare/Settings/Keybindings/KeybindingsConfiguration.swift
```diff
@@ -98,5 +91,8 @@ struct KeybindingsConfigurationView: View {
             addText: "Add",
             removeText: "Remove"
         )
+        .onReceive(NotificationCenter.default.publisher(for: .keybindsUpdated)) { _ in
```
mrkai77: Should be able to use `.onRecieve(.keybindsUpdated) {...}`

Loop/Utilities/TranslationLayer.swift
```diff
@@ -0,0 +1,93 @@
+//
```
mrkai77: Would recommend naming this `RectangleTranslationLayer` just in case we decide to implement more importers

Loop/Utilities/TranslationLayer.swift
```diff
@@ -0,0 +1,93 @@
...
+
+/// Represents the configuration of Rectangle app shortcuts.
+struct RectangleConfig: Codable {
+    let shortcuts: [String: RectangleShortcut]
+}
+
+/// Translates the RectangleConfig to an array of WindowActions for Loop.
+/// - Parameter rectangleConfig: The RectangleConfig instance to translate. (this works w/ both normal and pro)
+/// - Returns: An array of WindowAction instances corresponding to the RectangleConfig.
+func translateRectangleConfigToWindowActions(rectangleConfig: RectangleConfig) -> [WindowAction] {
```
mrkai77: Avoid top-level functions, probably makes sense to put all of these inside a `RectangleTranslationLayer` enum

Loop/Luminare/Loop/AdvancedConfiguration.swift
```diff
@@ -105,6 +105,13 @@ struct AdvancedConfigurationView: View {
             }
         }
 
+        LuminareSection {
+            Button("Import from Rectangle") {
```
mrkai77: "Import keybinds from Rectangle" might be more descriptive for the user

MrKai77 (changes requested):

SenpaiHunters:
In the most recent push, this optimises the whole 'WindowAction+Port', and while I was at it, I also made it fully update the UI, as that wasn't in the previous code. This PR should be ready to review now, later we can adjust some code and/or find better places for import button(s) etc.

Issue #111, opened 2023-12-03
✨ Hold down after delay instead of tapping.


JaceThings:
### Feature Request Description

I think it would be cool to allow holding down a key to cycle between sizes in a cycle. After a certain amount of time holding it, it would start cycling through the options of that particular key. It just feels a lot more intuitive than tapping repeatedly.

### Final Checks

- [X] My issue title is descriptive
- [X] This is a single feature request (multiple feature requests should be reported individually)

MrKai77:
Interesting, approach! Will look into this!

MrKai77:
So would Loop cycle through each cycle action after a user-set delay? Or did you have a type of delay in mind? (maybe there's something similar built-in to MacOS?)

JaceThings:
> So would Loop cycle through each cycle action after a user-set delay?

Pretty much! I thought about this over the weekend, and I thought the most "helpful" thing would be two timers:

1. Delay to hold to activate the "auto cycle."
2. "Decision Delay", which would be the amount of time the cycle appears on the screen for.

So this would be the Scenario.

Auto Cycle Delay: 5 Seconds
Decision Delay: 3 seconds

1. I hold down my activator keys. 
2. The first item (Action 0 for this example) in the cycle appears on the screen. 
3. I wait for 5 seconds. 
4. Action 1 then appears. 
5. I now wait for 3 seconds. 
6. Action 2 appears. 
7. I now wait for 3 seconds
8. ...

The "Decision Delay" is an extra feature I added to allow the user to avoid an initial delay when accessing action 0. If there was only one delay, pressing and holding the key would cycle through the actions too quickly when holding the action key down and wanting to activate the 0th action.

> (maybe there's something similar built-in to MacOS?)

Closest thing i can think of is when you hold down a key in macOS and you have "repeat letter" enabled.

MrKai77:
That makes a lot of sense, thanks!
I was also thinking that it would be hard to access the first action if there was only one delay, but having two would be much easier to use!
Anyways, this does seem like a more complicated feature compared to the other ones currently in Loop, so I think that it might be better to implement this after reorganizing/redesigning the settings window. (as this would probably be a more of a power-user feature)

JaceThings:
Yep, agreed. I would possibly add it to its own tab/section, depending on how it's structured. I will reply to the other settings design ticket soon; finishing work!

MrKai77:
Awesome, thank you so much!
[...]

PR #506, opened 2024-07-13
✨ Ability to advance cycles with a left click

MrKai77:
See here: https://github.com/MrKai77/Loop/issues/111#issuecomment-1931183399


SenpaiHunters (approved):
I accept these changes, as going forwards this will allow a custom radial menu!

PR #528, opened 2024-07-25
win11 style window drag [WIP]

FormalSnake:
As seen in the video, it is still very much WIP.
Since this is my first time implementing a feature in Loop, I am opening a PR already so that maintainers can tell me if it's worth to continue building, if i'm following good practices, etc.

https://github.com/user-attachments/assets/9270b970-3923-4134-be74-40bc2b8d1e08

There is a bug rn where it sometimes keeps showing itself for some reason.


FormalSnake:
Sorry, but starting Sunday I will be on vacation for 2 weeks. I'll see if i can do anything this saturday.

SenpaiHunters:
Thanks for the PR. However, we already have a working and active version locally; we just need to push the branch. Therefore, instead of using the incomplete code, we'll use our version.

FormalSnake:
yep I confirmed with kai, np.

PR #545, opened 2024-08-30
✨ Improved window action selector

MrKai77:


SenpaiHunters (approved):
Fixed a few things afterwards, we need to watch out in regards to changed `WindowDirections` if users indicate issues with movement not being correct. Otherwise, this is good to go.

PR #566, opened 2024-09-20
✨ Ability to use macOS 15's window manager with Loop

MrKai77:
This will allow the user to use the system window manager to move windows when available, and fallback to Loop's own window movement system (`WindowEngine`) when an action is more specialized.


SenpaiHunters (approved):
Love to see it, and it's so smooth! We may be able to do some code optimising on the `SystemWindowManager`!

Issue #462, opened 2024-06-25
🐞 Window doesnt move like the preview suggests


T1mVo:
### Bug Description

When moving a window via the keyboard, it does not move as the preview suggests.

### Steps To Reproduce

1. Bind the move actions to some keys
2. Move a window

### Expected Behavior

The window should move into the exact same position as the preview.

### Actual Behavior

The window only roughly moves to the position shown by the preview.

### Screenshots

https://github.com/MrKai77/Loop/assets/94936637/6766dd2a-b991-40a7-be14-44226d588db1

### MacOS Version

Sonoma 14.2

### Loop Version

Version 1.0.0 (1030)

### Final Checks

- [X] My issue is written in English
- [X] My issue title is descriptive
- [X] This is a single bug (multiple bugs should be reported individually)

MrKai77:
That is very interesting behavior. Do you have any external screens connected?

T1mVo:
Yes, the recording was made on a 32:9 monitor. Though the problem also exists on the built-in screen.

MrKai77:
Can you reproduce this bug when the external monitor is not plugged in?

T1mVo:
> Can you reproduce this bug when the external monitor is not plugged in?

Yes.

---

Some things i observed:

- The window goes too far into direction it got moved, e. g. If I move the window to the left, it overshoots the preview to the left.
- Only the last changed direction affects the prior issue, e. g. If I move the window to the top and then to the right, it only overshoots to the right, not to the top.

MrKai77:
Alright. Can you send me the file located at: `~/Library/Preferences/com.MrKai77.Loop.plist`?

T1mVo:
Sure, sent via email.

MrKai77:
Awesome, I will take a look! In the meanwhile, can you enable window animations in the Advanced tab, and see if Loop will behave differently compared to if it were off?

T1mVo:
Its actually on already. First thing I tried was to turn it off as it is still under development. Issue still exists though.

MrKai77:
I also just loaded your prefs into my Loop, and it seems to be working fine. Can you reproduce this with non-terminal-related apps as well?

T1mVo:
Yes, happens with other applications as well.

MrKai77:
Can you try running the latest [prerelease](https://github.com/MrKai77/Loop/releases/tag/prerelease)?

T1mVo:
Tried it, no change

[...]

PR #567, opened 2024-09-22
🐞 Fix window moving off-screen

T1mVo:
This commit fixes windows being moved off-screen via the keyboard shortcut when padding and animation are enabled.

The bug was caused by applying padding to a zero sized screenFrame. The bounds check in the animation would then fail.

Further explanation [here](https://github.com/MrKai77/Loop/issues/462#issuecomment-2365188359).

### DIFF
--- a/Loop/Window Management/WindowEngine.swift
+++ b/Loop/Window Management/WindowEngine.swift
@@ -94,15 +94,17 @@ enum WindowEngine {
             return
         }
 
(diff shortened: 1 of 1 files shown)


MrKai77:
Thank you! Is the comment have written [here](https://github.com/MrKai77/Loop/issues/462#issuecomment-2366776662) also something you plan to fix in this PR, or is this ready for review?

T1mVo:
This is ready for review.

MrKai77 (approved):
Looks good and makes sense, thanks!

PR #574, opened 2024-09-27
🐞 Fix window moving further than preview

### LINKED ISSUE #462: 🐞 Window doesnt move like the preview suggests

T1mVo:
This PR introduces a `isPreview` argument for the `WindowAction.getFrame()` function to differentiate between calculating the preview and the target window. If the final frame for the target window gets calculated, the last action will be ignored because `LoopManager.lastTargetFrame` is already at the desired size/position.

Closes #462

### DIFF
--- a/Loop/Preview Window/LuminarePreviewView.swift
+++ b/Loop/Preview Window/LuminarePreviewView.swift
@@ -62,7 +62,7 @@ struct LuminarePreviewView: View {
             .offset(x: actionRect.minX, y: actionRect.minY)
             .scaleEffect(CGSize(width: scale, height: scale))
             .onAppear {
-                actionRect = action.getFrame(window: nil, bounds: .init(origin: .zero, size: geo.size))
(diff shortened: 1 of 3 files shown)


MrKai77 (approved):
Fixed a small bug where this would stop working when the preview was disabled, but otherwise, looks good :)

PR #576, opened 2024-09-27
🐞 Fix crash when hiding or minimizing windows

### LINKED ISSUE #573: 🐞 Crash when minimizing or hiding an app with an keybind
### Bug Description

When having a keybind set to either hide, or minimize an app, Loop crashes instantly.

### Steps To Reproduce

1. Keybind to minimize or hide app
2. Use Keybind on any app

### Expected Behavior

To hide or minimize any app without crashing Loop.

### Actual Behavior

Loop crashed instantly without minimizing or hiding any app.

### Screenshots

https://github.com/user-attachments/assets/fb306147-5083-44d4-8af3-03075125976f

### macOS Version

Sequoia 15.0

### Loop Version

🧪 1.2.0 (1327)

### Additional Context

_No response_

### Final Checks

MrKai77:
This fixes a bug where Loop would crash when trying to minimize or hide a window. This is because there is no "frame" for Loop to calculate (even though it was trying to get one) when it was trying to calculate a target frame.

### DIFF
--- a/Loop/Window Management/WindowAction.swift
+++ b/Loop/Window Management/WindowAction.swift
@@ -134,7 +134,8 @@ struct WindowAction: Codable, Identifiable, Hashable, Equatable, Defaults.Serial
     }
 
     func getFrame(window: Window?, bounds: CGRect, disablePadding: Bool = false, screen: NSScreen? = nil) -> CGRect {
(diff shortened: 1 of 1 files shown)


SenpaiHunters (dismissed):
I forgot about loop being able to do this, anyway push her to prod!

PR #587, opened 2024-10-06
🐞 Fix custom percentage position

### LINKED ISSUE #581: 🐞 Percentage position not working as expected
### Bug Description

I've been playing with loop and love it.
I've set up some custom keybindings for my home setup based on position but want to more them the using percentages so that they scale appropriately when I connect to other screens

I've tested this with a single keybind which should occupy the lower right 6th of the screen (33% width, 50% height) but the percentage postioning isn't working as expected

### Steps To Reproduce

Create custom keybind with 50% height and 33% width, positioned at 67%, 50%
Screenshot 2024-10-01 at 08 56 44
Screenshot 2024-10-01 at 08 57 32

### Expected Behavior

To create the expected behaviour I have to set the position height to 31% rather than 50%
Screenshot 2024-10-01 at 08 57 54

### Actual Behavior

Keybind is positioned off the screen
Screenshot 2024-10-01 at 08 56 44

### Screenshots

_No response_

### macOS Version

Sonama 14.7

### Loop Version

Version 1.1.1 (1259)

### Additional Context

_No response_

### Final Checks

ShawnZhong:
Fix #581

### DIFF
--- a/Loop/Window Management/WindowAction.swift
+++ b/Loop/Window Management/WindowAction.swift
@@ -313,7 +313,7 @@ private extension WindowAction {
                 }
 
                 if let yPoint {
-                    result.origin.y += bounds.width * (yPoint / 100.0)
+                    result.origin.y += bounds.height * (yPoint / 100.0)
                 }
             }
         } else { // positionMode would be generic


MrKai77 (approved):
Thank you, I can't believe I missed this 🙃

PR #591, opened 2024-10-08
✨ Make window smaller/larger by scrolling

### LINKED ISSUE #521: ✨ Make window larger/smaller by scrolling up/down

Moving windows to different locations using only the mouse is a rather handy feature. However, another frequent operation is making the window larger or smaller, which is currently achievable only by keyboard shortcuts. I think it would be brilliant if we could do that by pressing the middle mouse button (to trigger loop) and then resize the window by scrolling, if that's possible. :)

### Screenshots

_No response_

### Additional Context

_No response_

### Final Checks

ShawnZhong:
Close #521

Untitled

I don't know what to do with scrolling left or right. They are currently set to make the window smaller / larger. 

An alternative could be:
- Scroll up: larger height
- Scroll down: smaller height
- Scroll right: larger width
- Scroll left: smaller width

### DIFF
--- a/Loop/Managers/LoopManager.swift
+++ b/Loop/Managers/LoopManager.swift
@@ -57,8 +57,8 @@ class LoopManager: ObservableObject {
 
         mouseMovedEventMonitor = NSEventMonitor(
             scope: .all,
-            eventMask: [.mouseMoved, .otherMouseDragged],
-            handler: mouseMoved(_:)
+            eventMask: [.mouseMoved, .otherMouseDragged, .scrollWheel],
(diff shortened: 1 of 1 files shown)


MrKai77:
Thanks so much for the PR! I've been holding off on this while I figure out how to make #138 work since the system APIs have made things trickier than expected. Once I get that sorted, I'll take another look at this :)

ShawnZhong:
Oh, I wasn't aware of the ongoing work on gesture support. Take your time on your implementation! I will close this PR for now. We can circle back once the gesture support is ready.

PR #595, opened 2024-10-14
🐞 Fix cycle actions not working

MrKai77:
Not sure how I didn't catch this bug earlier :/

This fixes a bug where cycle keybinds stop advancing after switching screens. Please test before approving, since again, I find it really odd that this bug originally even made it to production.


ShawnZhong:
I encountered this exact issue previously as well! I ultimately removed the screen switching from the Left/Right cycles. Will test if it is now working with this PR.

ShawnZhong:
Just tested that the issue is reproducible on 1.2.0 (1356) and fixed by this PR.

MrKai77:
@ShawnZhong awesome thank you so much for testing!!

PR #605, opened 2024-10-31
✨ Add Maximize Height & Width

### LINKED ISSUE #503: ✨ Max Height without Modifying Width

- Moved from Rectangle to Loop because of the GREAT window animation when resizing them! 
- There's a feature in Rectangle that I use every day that to be able to maximize the window size without modifying the width of it, many times I have Chrome max height but left the width untouched.
- I saw there are options to adjust the `with & height` in the `custom keybinds`, so the easier way to fulfill this feature request would be to give us `the option to un/check width or height`, so when changing width/height, the other property remains untouched.

### Screenshots

CleanShot_2024-07-10_11-38-09

### Additional Context

_No response_

### Final Checks

ShawnZhong:
Close #503


MrKai77:
I think there’s a more straightforward approach to achieving this feature. By adding one more `if` statement directly inside `calculateTargetFrame`, you could simplify the logic significantly (look at the `centering`/`undo`/`initial frame` logic as examples). Inside this statement, you could take the window's width and maximize the height, and set that as the `result`. Let me know if you need more details - but this should end up being a much simpler way of achieving the same functionality.

ShawnZhong:
> I think there’s a more straightforward approach to achieving this feature.

That completely makes sense. I was too obsessed with making it general.

MrKai77:
Yes, this seems perfect! It would be a good idea to add a maximize width as well. Once that’s in place, this should be complete!

MrKai77 (approved):
Just tested locally, works perfectly! Thank you so much!!

Ender-Wang:
> Just tested locally, works perfectly! Thank you so much!!

Do we have a plan for when v1.2.0 will be released? I saw so many new features merged. Cannot wait to try them.

Duckbilled:
I have been testing this and in my case it narrows the window a slight bit. I think it might be due to padding. Is this expected or an inconsistency?

I added a screen recording. Note the orange window behind it to see the change in width.

https://github.com/user-attachments/assets/795a6443-5e5a-4ad1-a5d7-71922c913386

ShawnZhong:
Thanks for the report @Duckbilled! I can see that the window is narrowed in your recording.  

I was trying to reproduce this but I couldn't so far. I also tried to change some settings and none of them seemed to make a difference:
- Preview -> Padding
- Behavior -> Respect Stage Manager
- General -> Use macOS window manager when available

I will keep an eye on this. If you have any other ideas, please let me know.

Duckbilled:
I have padding at 10px, but padding disappears when enabling `Use macOS window manager when available`. When it is enabled it maximises with the correct width, but my preferred padding is not used.

_Some of my settings:_

`Radial menu` &rarr; off
`Preview` &rarr; off
`Include padding` &rarr; `Simple, 10px`
`Respect stage manager` &rarr; off
`Disable cursor interaction` &rarr; off

Duckbilled:
I think the issue is, is that it assumes the width of the current window, and then adds another padding (+10px in my case). So basically when activating max height, it has +20px in horizontal padding. I don't have this issue when I completely turn off padding.

MrKai77:
@Duckbilled, if there is an issue you wish to report, please open a new issue :)

PR #631, opened 2024-12-13
🐞 Fix bug where frontmost application receives key events

zenangst:
This fix addresses the issue where pressing multiple keys quickly can cause a discrepancy and send keyboard events to the frontmost application.

I added a new instance variable called `lastKey` because `pressedKeys` is a `Set` which is not ordered.

Using the new instance variable as a fallback has fixed the issue for me.


MrKai77 (approved):
Works like a charm, thank you so much! Also, I didn't know about how it was possible to put in `#line` inside print statements, love it!

PR #656, opened 2024-12-31
🐞 Fix memory leak in `NSEventMonitor`

iSapozhnik:
| Before fix | After fix |
|---|---|
| SCR-20241231-toqh | SCR-20241231-ttce |


MrKai77 (approved):
Great catch, thanks for addressing this!!

iSapozhnik:
My pleasure, @MrKai77

PR #666, opened 2025-01-09
✨ Respect the window tiling spacing set in `com.apple.WindowManager`

zenangst:
In macOS Sequoia, you can tweak the spacing around windows. The key is `TiledWindowSpacing`.

I also bumped up the default from `8` to `9` (I think that's right based on my testing).

If you want to give it a shot, just run these commands.

```
defaults write com.apple.WindowManager EnableTiledWindowMargins -bool false; 
defaults write com.apple.WindowManager TiledWindowSpacing -float 16; 
defaults write com.apple.WindowManager EnableTiledWindowMargins -bool true;
```

**Note**
Toggling `EnableTiledWindowMargins` will move the windows you already have tiled. So, you don't need to restart the `WindowManager` to make the change.

### DIFF
--- a/Loop/Managers/SystemWindowManager.swift
+++ b/Loop/Managers/SystemWindowManager.swift
@@ -110,24 +110,21 @@ class SystemWindowManager {
             windowManagerDefaults?.bool(forKey: "EnableTilingByEdgeDrag") ?? false
         }
 
+        static var enablePadding: Bool {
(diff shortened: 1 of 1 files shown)


MrKai77 (approved):
Great find! I made the small change of separating `EnableTiledWindowMargins` and `TiledWindowSpacing`, just in case we make this configurable from Loop in the future, but otherwise looked perfect!

g7aaa (commented):
CGEventMonitor

PR #667, opened 2025-01-09
💫 Improve window animations

zenangst:
- Use the current refresh rate of the current display for `.frameRate`.
- Reduce the strain on the accessibility framework by never setting the same `position` or `size` twice. This should help reduce any redundant computation that might occur when the animation happens.
- Use `round` on `newFrame` so that we don't try and render frames with subpixel precision and make the compare statements work as intended.

### DIFF
--- a/Loop/Extensions/NSScreen+Extensions.swift
+++ b/Loop/Extensions/NSScreen+Extensions.swift
@@ -16,6 +16,17 @@ extension NSScreen {
         return deviceDescription[key] as? CGDirectDisplayID
     }
 
+    var displayMode: CGDisplayMode? {
+        guard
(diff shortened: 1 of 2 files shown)


zenangst:
Don't know if this is just me, but it does feel "smoother" with these changes. Could just be all in my head :)

Please take it for a spin to verify that the changes work as intended. I'm running them locally now.

MrKai77 (approved):
Tested locally, works well! I also moved the display mode code into the `NSScreen+Extensions` file. While I didn’t notice a visible difference, this is a good improvement as it also potentially reduces unnecessary refreshes and resizes. Thank you! :)

PR #668, opened 2025-01-09
💫 Improve animation settings together with `WindowEngine`

zenangst:
Respect animation settings configured in WindowManager's user defaults.


MrKai77 (approved):
Made a tweak so that `window.enhancedUserInterface` takes precedence over the system window manager’s settings. Other than that, this looks fantastic, thanks!

PR #675, opened 2025-01-21
🔧 Refactor key modifier handling in Keycorder

### LINKED ISSUE #669: 🐞 Ghost key added when assigning keybind command
### Bug Description

When trying to assign Keybinds to commands Loop adds function(?) key press without me using any other key. 

After using just arrow keys this is what I get. The same happens for other key like PgUp, but not for numeric keys. 
image

### Steps To Reproduce

1. Open Loop
2. Choose Kebinds
3. Add new Keybind (optional)
4. Change Keybind
5. Press desired key ->, <- for ex. 

### Expected Behavior

No additional ghost key should be added

### Actual Behavior

Ghost key was added

### Screenshots

image

### macOS Version

15.2 (24C101)

### Loop Version

Version 🧪 1.3.0 (1418)

### Additional Context

_No response_

### Final Checks

SenpaiHunters:
- Updated the logic to retrieve current modifiers by using a mapping of modifier flags to key codes.
- Filtered out trigger keys from the current modifiers to ensure they do not exceed the key limit.
- Fixes a bug where adding a trigger key with an arrow would cause the FN key to also be added.
- Renamed variables for clarity and improved code readability.


MrKai77 (approved):
Just tested locally, works great! Thanks!

PR #676, opened 2025-01-21
✨ Disable update checks via variables

### LINKED ISSUE #662: ✨ Option to disable update checks

I have a quick request: could you add an environment variable that completely disables update checks? This would make builds fully deterministic for package managers like Nix and help users in restricted or offline environments.

i.e. Fastlane uses `FASTLANE_SKIP_UPDATE_CHECK` set to `1`

### Screenshots

_No response_

### Additional Context

I maintain the Loop package on Nix (https://github.com/NixOS/nixpkgs/blob/nixos-unstable/pkgs/by-name/lo/loopwm/package.nix#L37)

### Final Checks

- [X] I can help with the improvement or development of this feature (tick if you can help)

SenpaiHunters:
- Introduced `updatesEnabled` class to first check for the `LOOP_SKIP_UPDATE_CHECK` environment variable, allowing users to skip update checks by setting this variable, otherwise, setting `Defaults` to control automatic updates.
- This allows for disabling automatic updates with `defaults write com.MrKai77.Loop updatesEnabled -bool false` and reset (to default of `true`) with `defaults delete com.MrKai77.Loop updatesEnabled`, otherwise, an environmental variable of `LOOP_SKIP_UPDATE_CHECK=1`
- Modified the initialization logic to only set up the update timer if updates are enabled and the environment variable / defaults are not set.
- Added an environment variable configuration in the Xcode scheme to facilitate testing and development without update checks.


MrKai77 (approved):
Made a few little changes, looks good!

PR #694, opened 2025-02-28
🐞 Fix memory leak inside `CGEventMonitor`

zenangst:
This should help fix the issue where things slow down after a period of usage as it no longer keeps previous event monitors around after looping as it resigned its usage via `LoopManager.closeLoop`.

Before, event monitor was passed as retained to itself, which quickly created a retain cycle. Also, I added proper cleanup by invalidating the mach port when the event monitor is about to go out of scope.

- Pass user info as unretained into the `callback` to avoid the mach port never being released.
- Invalidate the event tap inside `deinit`.


SenpaiHunters (approved):

MrKai77 (approved):
I think it makes sense to now merge this, as it seems to fix the issue based on the latest info from other commenters. Will release a new dev version containing this fix so that a larger audience can participate in testing!

PR #698, opened 2025-03-06
✨ Add URL scheme support

SenpaiHunters:
- Implemented URL command handling in AppDelegate
- Registered custom URL scheme 'loop' in Info.plist
- Added URLCommandHandler to process incoming URLs
- Configured Apple Event handling for URL interactions

### DIFF
--- a/Loop/Utilities/URLCommandHandler.swift
+++ b/Loop/Utilities/URLCommandHandler.swift
@@ -1,0 +1,412 @@
+//
+//  URLCommandHandler.swift
+//  Loop
+//
+//  Created by Kami on 06/03/2025.
+//
+
+/*
+ Loop URL Scheme Documentation
+ ===========================
+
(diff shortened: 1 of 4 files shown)


Loop/Utilities/URLCommandHandler.swift
```diff
@@ -0,0 +1,412 @@
...
+
+            logWindowDetails(win, app, isLoop, isRegular, isVisible, false)
+
+            return !isLoop && isRegular && isVisible
+        }
+
+        print("[URLHandler] Found \(visibleWindows.count) eligible windows")
+
+        // Find target window
+        guard let window = findTargetWindow(from: visibleWindows) else {
```
mrkai77: I think it makes more sense to find the window with `WindowEngine.getTargetWindow()`?

MrKai77 (approved):
Seems good, will release a new development build with these changes!

PR #729, opened 2025-05-16
✨ Enable reverse cycling of window actions with Shift key

### LINKED ISSUE #696: ✨ Modifier key reverses cycle direction

It would be very useful to me to be able to move backward through one of the cycles I've configured. For any cycles longer than a few entries, it can get tedious having to loop all the way back around if you've missed your intended location. Much like `Shift+⌘+Tab` cycles in the reverse direction of `⌘+Tab` through active applications, can we have (something like) `Shift+<Trigger Key>+<Keybind>`?

I searched for similar feature requests but didn't find one. Apologies if this is a duplicate.

### Screenshots

_No response_

### Additional Context

_No response_

### Final Checks

guillaumecledat:
This is an initial implementation for [#696](https://github.com/MrKai77/Loop/issues/696), adding support for cycling backwards through actions when the Shift key is pressed.

With this code, backward cycling is enabled for all Cycle actions **if** the Shift key is pressed and neither the trigger key nor the keybind already includes Shift.

To support this behavior, I introduced an `isShiftPressed()` method in `KeybindMonitor`. It's a bit of a rough solution, so if you have a cleaner approach in mind, I'm happy to revise it.

Also, we should discuss the feature's scope. It might make sense to add a global toggle to enable/disable backward cycling *or* add per-cycle configuration of this behavior

Feedback is welcome—especially on implementation design and configurability.

### DIFF
--- a/Loop.xcodeproj/project.pbxproj
+++ b/Loop.xcodeproj/project.pbxproj
@@ -89,6 +89,7 @@
 		A8F0125B2AEDD7660017307F /* WindowAction.swift in Sources */ = {isa = PBXBuildFile; fileRef = A8F0125A2AEDD7660017307F /* WindowAction.swift */; };
 		A8F1E9662C253F5B00AAF871 /* Localizable.xcstrings in Resources */ = {isa = PBXBuildFile; fileRef = A8F1E9652C253F5B00AAF871 /* Localizable.xcstrings */; };
 		A8F1E9692C253F8D00AAF871 /* InfoPlist.strings in Resources */ = {isa = PBXBuildFile; fileRef = A8F1E9672C253F8D00AAF871 /* InfoPlist.strings */; };
+		F0642DAA2DD8C9F00049F167 /* CycleBackwardsView.swift in Sources */ = {isa = PBXBuildFile; fileRef = F0642DA92DD8C9F00049F167 /* CycleBackwardsView.swift */; };
 /* End PBXBuildFile section */
 
 /* Begin PBXFileReference section */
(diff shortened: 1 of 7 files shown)


MrKai77:
@Morniak actually, this looks great so far! You mentioned `isShiftPressed()` feeling like a rough solution, but I don’t think it is :)
Since `KeybindMonitor` was designed to track pressed keys during Loop’s activation, it totally makes sense to use it here.

As for the feature’s scope, I think it should be a global setting. Making it per-cycle might overcomplicate things for the average user.

guillaumecledat:
Great, thanks! Should I go ahead and add a global toggle to disable this feature?
If so, do you have a preferred location for it? I was thinking the Keybinds tab, under the Settings section, might be a good fit.

MrKai77:
Yes, that makes sense. But It would also make sense to make it it's own little "section", without a title, right below the trigger settings. And, do you think it would make sense to only show it if the keybinds list contains a cycle action as well?

guillaumecledat:
A new section below the trigger key makes sense, and showing it only when there's a cycle action is a good idea — I'm on it.
If the trigger key uses Shift, I think we should keep the toggle visible but disabled, and add a LuminareInfoView to explain why.

guillaumecledat:
The change has been pushed, you can check it out now @MrKai77.
Note: the section is toggled off, not disabled, when the trigger key includes Shift because LuminareInfoView doesn’t appear for disabled LuminareToggle.

SenpaiHunters (approved):

PR #734, opened 2025-05-25
✨ Stash windows on screen edge

### LINKED ISSUE #681: ✨ Stash windows on screen edge

Rectangle Pro has a feature that lets you stash windows on the screen edge (similar to an iPad). Since I only use my laptop screen, I find this feature immensely useful, as I can stash two Chrome windows while keeping a VS Code window open on the screen, and have a reference while coding. Although it has some bugs, particularly with "stay on top" enabled windows, I would love to see this feature in Loop.

Image

### Screenshots

_No response_

### Additional Context

_No response_

### Final Checks

guillaumecledat:
Implementing the "stash" feature discussed here https://github.com/MrKai77/Loop/issues/681.

As mentioned in the issue, this pull request aims to create an proof of concept to evaluate whether the core functionalities are solid enough to be included in Loop. In my opinion, the two main obstacles at the moment are:
- When stashed, the window remains focused.  
- When the window slides back onto the screen, it doesn't always appear as the topmost window.


Loop.xcodeproj/project.pbxproj
```diff
@@ -847,7 +852,7 @@
 				COMBINE_HIDPI_IMAGES = YES;
 				DEAD_CODE_STRIPPING = YES;
 				DEVELOPMENT_ASSET_PATHS = "\"Loop/Preview Content\"";
-				DEVELOPMENT_TEAM = 5F967GYF84;
```
SenpaiHunters: This needs to be changed back to the original signature

Loop/Managers/StashManager.swift
```diff
@@ -0,0 +1,220 @@
...
+        mouseMonitor = nil
+    }
+
+    /// Handles mouse movement events to reveal or hide stashed windows.
+    private func handleMouseMoved(_: NSEvent) {
+        // TODO: - Test with multiple displays
+        for (windowID, window) in stashedWindows {
+            let screen = window.screen
+
+            // TODO: Investigate why we need to flip mouse Y coordinate
```
SenpaiHunters: Consider adding fallback behaviors (e.g., log failure, retry, focus another window).
Ensure you've tested edge cases like full-screen apps or hidden spaces.

Hardcoded magic coordinate flipping can easily break on non-standard screen arrangements. Add a helper method with a documented rationale, e.g.: 

```swift
extension CGPoint {
    static func flippedY(forScreen screen: NSScreen, point: CGPoint) -> CGPoint {
        return CGPoint(x: point.x, y: screen.frame.maxY - point.y)
    }
}
```

Loop/Managers/StashManager.swift
```diff
@@ -0,0 +1,220 @@
...
+    }
+}
+
+// MARK: - Mouse moved listener
+
+private extension StashManager {
+    private func startListeningMouseMoved() {
+        print("Listening for mouse moved events…")
+
+        mouseMonitor = NSEvent.addGlobalMonitorForEvents(matching: .mouseMoved) { [weak self] event in
```
SenpaiHunters: If `stopListeningMouseMoved()` is never called (e.g., crash, logic bug), it could cause lingering global event listeners, causing a memory leak, `deinit` this if not needed, or add fallbacks just in case.

No throttling of mouse events → performance risk - Use a throttling mechanism, e.g., debounce with a short timer (e.g., 100ms):
```swift
private var mouseMoveWorkItem: DispatchWorkItem?

private func handleMouseMoved(_ event: NSEvent) {
    mouseMoveWorkItem?.cancel()
    let workItem = DispatchWorkItem { [weak self] in self?.processMouseMovement(event) }
    mouseMoveWorkItem = workItem
    DispatchQueue.main.asyncAfter(deadline: .now() + 0.1, execute: workItem)
}
```

Loop/Managers/StashManager.swift
```diff
@@ -0,0 +1,220 @@
...
+            let invertedMouseY = screen.frame.maxY - NSEvent.mouseLocation.y
+            let mouseLocation = CGPoint(x: NSEvent.mouseLocation.x, y: invertedMouseY)
+
+            let isWindowRevealed = revealedWindows.contains(windowID)
+            let isMouseOverStashedWindow = NSPointInRect(mouseLocation, window.stashedFrame)
+            let isMouseOverRevealFrame = NSPointInRect(mouseLocation, window.revealFrame)
+
+            if isWindowRevealed, !isMouseOverRevealFrame, !isMouseOverStashedWindow {
+                hideWindow(window, animate: animateHide)
+            } else if isMouseOverStashedWindow {
```
SenpaiHunters: Overlap detection may produce jitter or flicker if `revealFrame` and `stashedFrame` are very close. Add a debounce or timer delay to reveal/hide logic. Also document overlapping behavior.

Loop/Managers/StashManager.swift
```diff
@@ -0,0 +1,220 @@
...
+    private func stash(window: Window, on screen: NSScreen, action: WindowAction) {
+        let windowID = window.cgWindowID
+        let currentFrame = window.frame
+        var revealFrame = currentFrame
+        var stashedFrame = currentFrame
+        var direction: StashDirection
+
+        if action.direction == .stashRight {
+            revealFrame.origin.x = screen.frame.maxX - revealFrame.width - Defaults[.padding].right
+            stashedFrame.origin.x = screen.frame.maxX - stashedWindowPadding
```
SenpaiHunters: No bounds checking (e.g., frame width could cause window to bleed onto another screen or off-screen).  Clamp frame origins to screen bounds using a helper, I think there should be a bounds extension somewhere, if not, one can be made in this file.

Loop/Managers/StashManager.swift
```diff
@@ -0,0 +1,220 @@
...
+    private let animateReveal: Bool = true
+
+    // TODO: Store this in Defaults
+    private let animateHide: Bool = true
+
+    // TODO: Store this in Defaults
+    /// How many pixels of the window should be visible when stashed
+    private let stashedWindowPadding: CGFloat = 20
+
+    private var stashedWindows: [CGWindowID: StashedWindow] = [:]
```
SenpaiHunters: `stashedWindows` and `revealedWindows` are stored in memory only. If the app restarts or crashes, all stash state is lost. Persist at least the window IDs + direction + screen ID (not frames, which are volatile), this can be stored in `Defaults`.

Loop/Managers/StashManager.swift
```diff
@@ -0,0 +1,220 @@
+//
+//  StashManager.swift
+//  Loop
+//
+//  Created by Guillaume Clédat on 22/05/2025.
```
SenpaiHunters: No edge-case handling for window resizing
If the window is resized manually while stashed, the stashed or reveal frame could become invalid — e.g., reveal no longer shows content. Hopefully this won't happen, but from what I know of users, they may bump the resizing of the windows, which can cause issues down the road. Recalculate `stashedFrame` / `revealFrame` if the window size changes while stashed.

Potential overlap for multiple stashed windows
If two windows are stashed on the same screen and side, their `stashedFrame` may collide. Introduce offset stacking (like a cascading UI):
```swift
stashedFrame.origin.y -= CGFloat(index) * offsetPerWindow
```
Or, prevent stashing if a window is already stashed on that edge.

SenpaiHunters:
Really good commit, I like the feature was speaking to Kai about this a few days ago actually, so awesome timing, only a few things which could be taken care of before merging, I've added comments on them, and if you need help, feel free to reach out. I tested this locally and it works well! One last thing we'll need to adjust the readme to say we now support screen stashing!!

Loop/Managers/StashManager.swift
```diff
@@ -0,0 +1,220 @@
...
+    let revealFrame: CGRect
+    let stashedFrame: CGRect
+}
+
+class StashManager {
+    // TODO: Store this in Defaults
+    private let animateReveal: Bool = true
+
+    // TODO: Store this in Defaults
+    private let animateHide: Bool = true
```
mrkai77: I think `animateReveal` and `animateHide` should be combined into a single setting. But that's just my opinion, let me know if you've had a use case that required both settings to be separately configured.

Loop/Managers/StashManager.swift
```diff
@@ -0,0 +1,220 @@
...
+
+private extension StashManager {
+    /// Reveals a stashed window by moving it to its reveal frame.
+    private func revealWindow(_ window: StashedWindow, animate: Bool) {
+        let windowID = window.window.cgWindowID
+
+        guard !revealedWindows.contains(windowID) else { return }
+
+        // TODO: - Investigate activate not always working
+        window.window.activate()
```
mrkai77: This was an issue I encountered in one of my other projects as well. This can be resolved by using SkyLight APIs (which I'm fine with).

Take a look at: https://github.com/robwakefield/Amethyst/blob/e6f7a5d63e8829dae6d88ed4a4c3dc5c60158672/Amethyst/Model/Window.swift#L290

The `activate` function of a `Window`, should also call this (we can call it `focus` too), which should be contained within a separate function :)

Let me know if you need help with this one!

Loop/Managers/StashManager.swift
mrkai77: Use `NSEventMonitor`. This will also take care of cleanup.

Loop/Managers/StashManager.swift
```diff
@@ -0,0 +1,220 @@
...
+                revealWindow(window, animate: animateReveal)
+            }
+        }
+    }
+}
+
+// MARK: - Helpers
+
+private extension NSScreen {
+    func isSameScreen(_ other: NSScreen) -> Bool {
```
mrkai77: Move this into `NSScreen+Extensions`, and use `NSScreen.displayID` to get screen numbers instead :)

Loop/Managers/StashManager.swift
```diff
@@ -0,0 +1,220 @@
...
+            let screen = window.screen
+
+            // TODO: Investigate why we need to flip mouse Y coordinate
+            // I'm not exactly sure why we need to flip the mouse Y coordinate here, but it seems to work
+            let invertedMouseY = screen.frame.maxY - NSEvent.mouseLocation.y
+            let mouseLocation = CGPoint(x: NSEvent.mouseLocation.x, y: invertedMouseY)
+
+            let isWindowRevealed = revealedWindows.contains(windowID)
+            let isMouseOverStashedWindow = NSPointInRect(mouseLocation, window.stashedFrame)
+            let isMouseOverRevealFrame = NSPointInRect(mouseLocation, window.revealFrame)
```
mrkai77:
```suggestion
            let isMouseOverRevealFrame = window.revealFrame.contains(mouseLocation)
```

Loop/Managers/StashManager.swift
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+    /// Handles mouse movement events to reveal or hide stashed windows.
+    private func handleMouseMoved(_: NSEvent) {
+        // TODO: - Test with multiple displays
+        for (windowID, window) in stashedWindows {
+            let screen = window.screen
+
+            // TODO: Investigate why we need to flip mouse Y coordinate
+            // I'm not exactly sure why we need to flip the mouse Y coordinate here, but it seems to work
```
mrkai77: A y-axis flip is required as we are converting to CoreGraphics coordinate space (origin @ top left) from AppKit coordinate space (origin @ bottom left). Rather than manually flipping the coordinate here, look into the `CGRect.flipY(_:)` function within `CGGeometry+Extensions.swift`, where you should be passing in `NSScreen.screens[0]` as a reference point to flip Y against.

MrKai77 (changes requested):
Overall, this is great code, really well-architected. I’ve left a few comments in places, but nothing too major. Excited to see this feature fully functional! :)

SenpaiHunters (changes requested):

Loop/Managers/StashManager.swift
mrkai77: Note: Use method in `CGGeometry+Extensions` where you should pass in the first screen in `NSScreen.screens` :)

guillaumecledat:
Hey, thanks for the feedback! I tried your solution using the SkyLight API, and it’s working like a charm so that's great news. Would you like me to make one commit per change, one commit per review, or one commit for both your reviews?

SenpaiHunters:
Whatever you feel is better for you working, i recommend smaller changes per commit so you can keep a track of what's changed per edit and not just one massive mess. You can group up multiple changes at once, but for more confusing things, starting fresh is nice.

Loop/Managers/StashManager.swift
guillaumecledat: One setting is good enough.
Fixed in https://github.com/MrKai77/Loop/pull/734/commits/f1d21a401ff97a16a38b9eb8ebc2e4d95f93e13f.

Loop/Managers/StashManager.swift
guillaumecledat: Fixed in https://github.com/MrKai77/Loop/pull/734/commits/ecbfcc76ec5db391a4224417f520ef14b618305c.

Loop/Managers/StashManager.swift
guillaumecledat: Fixed in https://github.com/MrKai77/Loop/pull/734/commits/33a6b47d2758de934b23aede700f1d65e5bd74ae.

Loop/Managers/StashManager.swift
guillaumecledat: Fixed in https://github.com/MrKai77/Loop/pull/734/commits/f71c8e31854f169bd28a31b4df4196150c61129e.

Loop/Managers/StashManager.swift
guillaumecledat: Fixed in https://github.com/MrKai77/Loop/pull/734/commits/91c217cb34f18172711ea0fb10334b3bdb2f0db9.

Loop/Managers/StashManager.swift
guillaumecledat: Thank for the explanations!
Fixed in https://github.com/MrKai77/Loop/pull/734/commits/43c75d12059b24f68637b5e3edcce4ef0733dddd.

Loop.xcodeproj/project.pbxproj
guillaumecledat: Fixed in https://github.com/MrKai77/Loop/pull/734/commits/578bcb81a91a166a4b5a4cf19c291a8e2e6f3478.

Loop/Managers/StashManager.swift
guillaumecledat: Fixed in https://github.com/MrKai77/Loop/pull/734/commits/e9574489173ed2eee3be79419ea6f474e57aa20c.
I used a debounce interval of 50 ms because 100 ms felt slightly sluggish.

Loop/Managers/StashManager.swift
guillaumecledat: A throttle for the reveal/hide as been added in https://github.com/MrKai77/Loop/pull/734/commits/7049f98f0be820de7b585f7a794c6db46dbbcc3e.
I will add documentation in a future commit (Let's keep this conversation open until then).

Loop/Managers/StashManager.swift
guillaumecledat: You make a great point, but I’m not quite sure how to approach this. When restoring the `stashedWindows` from stored data, I’ll eventually need their frames. I ran a test using `WindowEngine.windowList` to find the corresponding window, which would involve figuring out whether the window’s frame is `revealFrame` or `stashedFrame`, and then computing the other value. This seems feasible although it need a bit more work. 

However, as with the `unfocus` method issue, macOS Mission Control’s "Spaces" make this much more complicated. Specifically, if a stashed window resides on a different Space than the current one, `WindowEngine.windowList` won’t list it.

Should I continue working on my implementation, even though it would be a best-effort solution that might occasionally fail?

Loop/Managers/StashManager.swift
guillaumecledat: Regarding the overlap behavior, I really like the cascading UI idea, but I think it might require significant rework. Since I have many other tasks to handle, I’m considering an approach where you can only stash one window per edge, per screen. If the user attempts to stash a window in an already occupied position, it will replace the existing stashed window.

The cascading idea could be implemented either once the rest of the work is complete or in a later pull request.

Another option would be to keep the "one window per stash location" logic but introduce more `StashDirection` values, like `topHalfRight` or `bottomThirdRight`. This way, multiple windows could be stashed on the same edge, as long as their areas don’t overlap.

Edit: I also think the cascading UI will conflict with the fact that we make the stashed window under the mouse the frontmost window when revealing it.

SenpaiHunters:
Just as a reminder, when adding something, refer to the [contributing](https://github.com/MrKai77/Loop/blob/develop/CONTRIBUTING.md) guide, as it outlines
```sh
# If you're committing, you must use the following emojis at the start:
# 🐞 Bug fixes must include a bug emoji.
# ✨ Added features must include a star emoji.
# 🌐 Localization must include a globe emoji.
# An example of this would be:
# git commit -m "✨ Add wallpaper theming"
```

Loop/Managers/StashManager.swift
SenpaiHunters: It's more of an if loop crashes it will break, so it adds fallbacks. It will require features fallbacking, than on first launch, validate if we had any stored windows in memory, if we did, open them in it's spots, if not, proceed as normal. 

If it's too hard for now, we can iterate on it in later revisions, so no stress.

Loop/Managers/StashManager.swift
SenpaiHunters: As i outlined in my other comment below, if it's currently too much, feel free to skip it until a later iteration.

guillaumecledat:
Do you want me to rebase to add missing emojis to commit messages? Also, in case of a refactoring, what emoji should I use?

SenpaiHunters:
For your final commit add `✨`and any changes which is a feature `✨`. You can play around with fun ones, like `🔧` or `🗑️` for "fixes" or "refactoring/optimisations"

guillaumecledat:
@MrKai77 @SenpaiHunters I pushed a major refactor of the feature.

The version you reviewed was more of a prototype than anything else, and my implementation was cutting a lot of corners.  
With this new version, many of your review points have been implemented and new features have been added.

This version includes:
- A working preview of stash actions
- Undo now works with stashed windows
- Overlap check logic
- Functional and configurable unfocus in the options
- Options added to the UI
- A new `WindowAction`, `.customStash` working like `.custom`
- A window restoration system (in case of crash)
- Detection of an attempt to stash between two monitors (although this is open for discussion, see below)

A few design choices I made and would like to discuss:
- When the user uses the keybind to unstash or when the application exits normally, stashed windows are moved at the center of the screen. I could have used `.initialFrame`, but when cycles are used, the value of `initialFrame` didn't seem consistent.
- There is no keybind to "reveal" a stashed window, so the feature only works with the mouse. I was considering adding a keybind to reveal the stashed window(s) on the focused screen, but I felt that adding a keybind that affects a window different from the one in focus might contradict Loop's existing logic.
- It's not possible to stash a window on the edge of the screen that leads to a second screen. Since the stashed window is mostly off-screen, the system considers it to be on the second screen. To prevent this, I check and block the action. However, currently this fails silently without notifying the user, which is not ideal. I see a few solutions for this:
    - If the user tries to stash to the right of their screen but there’s a second screen on the right, we stash the window to the right on the second screen.
    - We block the action and notify the user (via a notification, window wiggle, etc.)
- For `WindowDirection.customStash`, I reused the UI used by `WindowDirection.custom` with a few modifications. Let me know if these changes work for you and/or if you'd prefer me to create a second independent view to avoid the few if/else I added in `CustomActionConfigurationView`.

Some technical issues:
- `Window.cgWindowID` is not unique for two windows of the same process (e.g. two Finder windows). This causes unexpected behaviors when two such windows are stashed. I tried to find a solution, but nothing I tested worked.
- The behavior of the `smaller`, `grow`, `shrink` actions is weird with stashed windows, with an invalid window width.
- Using `smaller` on a window almost completely off-screen causes a crash. I’ll create a separate ticket for this.

What's left to do:
- Update the README
- Add icons for stash actions
- Anything you find during your review

Here is a keybind file I used while working on the feature:
[Loop Keybinds Stash.json](https://github.com/user-attachments/files/20553154/Loop.Keybinds.Stash.json)

Let me know what you think 🫡

MrKai77:
Thank you for all the updates and the detailed summary @Morniak!
I’ll be able to review everything this weekend and will share my feedback then :)

SenpaiHunters:
I've reviewed your PR (the code side of it) and see a lot of good changes since my last review. I will make a new build over the weekend to fully test, and so will Kai, but it looks good! Thank you for the extensive code comments.

Discussions:
1. Interesting; there could be something in the main address conflicting, or some values are being mixed up. This will require some debugging or creating a new function if nothing else works, as separate is nice.
2. Personally, if it's a feature added by us, adding a keybind to unhide it is fine. Let's see what Kai thinks.
3. I think the second option would be the easiest and cleanest, so we can treat all connected screens as "one" wide screen. I believe we'd need to do the same for the left if that is configured to be the left screen facing rather than the right.
4. That's fine to reuse it; no need to make a whole new component.

Technical issues:

1. Non-unique `cgWindowID` - This is a known limitation.

2. Resize Actions on Stashed Windows
Consider:
   - Temporarily moving the window on-screen for resize operations
   - Calculating resize operations based on the window's "logical" position rather than the actual position
   - Disabling or modifying resize behavior for stashed windows

3. Crash with `smaller` Action
We'll want to look into this before releasing the code, as it's a core issue. I don't think a ticket would be needed here, purely because the stashing is causing the problem. Adding a code breakpoint to find where it's crashed could be useful. We'll need to investigate this further before it's launched into production.

Loop/Extensions/Notification+Extensions.swift
```diff
@@ -10,6 +10,7 @@ import Foundation
 extension Notification.Name {
     static let updateBackendDirection = Notification.Name("updateBackendDirection")
     static let updateUIDirection = Notification.Name("updateUIDirection")
+    static let UIDirectionUpdated = Notification.Name("UIDirectionUpdated")
```
mrkai77: Improve naming. This is easily confusable with `updateUIDirection`, maybe `windowDidResize` or `windowResized` would be clearer.

Loop/Luminare/Settings/Keybindings/CustomActionConfigurationView.swift
```diff
@@ -29,13 +29,25 @@ struct CustomActionConfigurationView: View {
         }
     }
```
mrkai77: The `size` case should be called `expanded size` or `unstashed size` (on the UI) or similar for a stash keybind, otherwise it is hard to understand what size it's referring to.

Loop/Managers/StashManager/StashManager.swift
```diff
@@ -0,0 +1,462 @@
...
+
+    /// The time interval to debounce mouse moved events to avoid excessive processing.
+    private let mouseMovedDebounceInterval: TimeInterval = 0.05
+
+    /// The throttle interval for revealing/hiding windows when the mouse moves.
+    private let revealThrottleInterval: TimeInterval = 0.1
+
+    /// Two windows can be stacked along the same edge of the screen as long as there is enough non-overlapping space
+    /// to allow the user to easily position the cursor over either window.
+    private let minimunVisibleHeightToKeepWindowStacked: CGFloat = 100
```
mrkai77:
```suggestion
    private let minimumVisibleHeightToKeepWindowStacked: CGFloat = 100
```

Loop/Window Management/Window.swift
```diff
@@ -177,10 +232,39 @@ class Window {
         fullscreen = !fullscreen
     }
 
+    /// Check with the `NSRunningApplication` if the app is hidden (⌘H).
     var isHidden: Bool {
```
mrkai77:
```suggestion
    var isApplicationHidden: Bool {
```
This will help distinguish it from `isWindowHidden`.

Loop/Window Management/Window.swift
```diff
@@ -344,3 +428,10 @@ class Window {
         }
     }
 }
+
+extension Window: CustomDebugStringConvertible {
+    var debugDescription: String {
+        let name = nsRunningApplication?.localizedName ?? title ?? "<unknown>"
+        return "Window{id:\(cgWindowID),name:\(name)}"
```
mrkai77:
```suggestion
        return "Window(id:\(cgWindowID), name:\(name))"
```

Loop/Window Management/WindowDirection.swift
```diff
@@ -51,6 +51,19 @@ enum WindowDirection: String, CaseIterable, Identifiable, Codable {
...
+
+    // Stash (Quarters)
+    case stashTopLeftQuarter = "StashTopLeftQuarter", stashTopRightQuarter = "StashTopRightQuarter"
+    case stashBottomRightQuarter = "StashBottomRightQuarter", stashBottomLeftQuarter = "StashBottomLeftQuarter"
+
+    // Stash (Thirds)
+    case stashRightThird = "StashRightThird", stashLeftThird = "StashLeftThird"
+    case stashRightTwoThirds = "StashRightTwoThirds", stashLeftTwoThirds = "StashLeftTwoThirds"
+    case customStash = "Custom Stash"
+    case unstash = "Unstash"
```
mrkai77, line 56: Stashing in Loop should ideally be handled as a single action, rather than having separate actions for each possible unstashed frame. This approach would minimize confusion. For example, “right third stash” isn’t intuitive, since it actually refers to the unstashed window size, not the stashed state.

In short, I believe it makes more sense to rename ‎⁠`customStash⁠ to` simply `‎⁠stash⁠,` and to remove the other stash-related options. This change would also streamline the keybinds action picker and make it less cluttered.

cc. @SenpaiHunters

Loop/Window Management/WindowDirection.swift
```diff
@@ -73,8 +93,10 @@ enum WindowDirection: String, CaseIterable, Identifiable, Codable {
     var willShrink: Bool { WindowDirection.shrink.contains(self) }
     var willGrow: Bool { WindowDirection.grow.contains(self) }
     var willMove: Bool { WindowDirection.move.contains(self) }
+    var willStash: Bool { WindowDirection.stash.contains(self) }
     var willMaximize: Bool { [.fullscreen, .maximize, .almostMaximize, .maximizeHeight, .maximizeWidth].contains(self) }
     var willCenter: Bool { [.center, .macOSCenter, .verticalCenterHalf, .horizontalCenterHalf].contains(self) }
+    var isCustom: Bool { [.custom, .customStash].contains(self) }
```
mrkai77:
```suggestion
    var isCustomizable: Bool { [.custom, .customStash].contains(self) }
```

MrKai77 (changes requested):
Here's my review. Everything has definitely improved, and you've made very good progress!

As for your discussion points:
1. Sure, this works as a temporary solution, but I think it makes more sense to investigate the cause of this inconsistency. It would definitely be more ideal to use `.initialFrame` rather than centering everything in my opinion.
2. Actually, I have an idea. How about for each stash action, only one window could be associated? Of course, the user could make multiple keybinds for the same stash area, but in practice, each action (or keybind) should only be in charge of a single window at any given time. That way, the same keybind could be used to _toggle_ the stashed/unstashed state of a window. And then to "unlink" a window from a stash action, all the user would need to do would be to apply another loop action on it, or to drag it manually. But of course, this is just an idea I had right now, and would like to discuss this further. This would also prevent the need for a dedicated "reveal" action, which, based on my understanding, would reveal all stashed windows at once?
3. As @SenpaiHunters states, I also think that it would be ideal to treat all screens as one "wide" screen.
4. It looks good! I also added a comment on its implementation in my review.

As for the technical issues:
1. Yes, this is a known issue, and I've been experimenting with potential fixes. For this PR, it will be fine to not address this :)
2. Agree with @SenpaiHunters said
3. A ticket isn't needed for a bug that doesn't exist in the current version of the app on the `develop` branch. It will be best to resolve this bug before merging.

Anyway, that is my review! Again, you've made really good progress, and it works really well on my system already :D

Loop/Window Management/WindowDirection.swift
SenpaiHunters: I agree on this; it's not a massive thing, but it's cleaner code-wise. Keeping it simple and in one case is always nicer logically than having it in many, such as when it's in thirds.

SenpaiHunters:
Okay, looking at the crash logs, I can see the issue is in the `PreviewController.setAction` method where it's trying to set a window frame with invalid coordinates (infinity and negative infinity values). I have added some code, and it no longer seems to crash. CC @Morniak and CC @MrKai77 if either of you would like to see how I fixed it, or suggest something better, feel free. This is just a quick and dirty fix.

---

Files to fix:

At the end of the `extension CGRect {` extension in `CGGeometry+Extensions.swift`, add the following code
```swift:CGGeometry+Extensions.swift
	/// Returns true if the rectangle is finite, false otherwise.
    var isFinite: Bool {
        origin.x.isFinite &&
            origin.y.isFinite &&
            size.width.isFinite &&
            size.height.isFinite &&
            !isNull &&
            !isInfinite
    }
```

In the function `func setAction(to action: WindowAction) {` at `PreviewController.swift`, you need to add the following
```swift:PreviewController.swift
    func setAction(to action: WindowAction) {
        guard
            let windowController = controller,
            let screen,
            !action.direction.willChangeScreen,
            action.direction != .cycle
        else {
            return
        }

        /// Check screen bounds
        print("Screen frame: \(screen.frame)")
        print("Screen safeScreenFrame: \(screen.safeScreenFrame)")

        // Validate screen bounds before proceeding
        guard screen.safeScreenFrame.isFinite else {
            print("ERROR: Invalid screen bounds detected!")
            return
        }

        let targetWindowFrame = action.getFrame(
            window: window,
            bounds: screen.safeScreenFrame,
            screen: screen,
            isPreview: true
        )
        .flipY(maxY: NSScreen.screens[0].frame.maxY)

        // What is the screen's frame
        print("Target frame: \(targetWindowFrame)")

        // Validate target frame before setting
        guard targetWindowFrame.isFinite else {
            print("ERROR: Invalid target frame calculated!")
            return
        }

        let isCurrentlyTransparent = windowController.window?.alphaValue == 0
        let shouldBecomeTransparent = targetWindowFrame.size.area == 0
```
This has the side effect of no longer crashing, but putting the window on the screen again, in the size of a `rect`, it will go smaller until the window itself cannot go any smaller, because that's how loop functions work.

guillaumecledat:
Thanks for the review, I’ll take a look.

Regarding the crash: my message should have been clearer, but it also happens on develop (just manually move a window almost completely off-screen).
@SenpaiHunters I’ve tested your code, and it does fix the crash. Should I include the fix in this PR, or, since the issue also occurs on develop, would it be better to open a separate PR for it?

SenpaiHunters:
Hey, I thought I replied to this. I must have forgotten, as it is caused by this PR. The code should be included in said PR, not as a separate fix or separate PR.

guillaumecledat:
1. I couldn't find the old problems when unstashing using the first frame, so I added that change. Let me know if you spot some weird behavior. 
2. Upon further thought, I think that my original idea of an action to show all stashed windows simultaneously could be awkward. Your solution addresses this but it does introduce some difficulty when using `cycle` with different `stash` positions.
3. I pushed the change.
4. Discussed in my preview comment

Loop/Extensions/NSScreen+Extensions.swift
```diff
@@ -127,4 +131,12 @@ extension NSScreen {
             return CGSize(width: 72.0, height: 72.0)
         }
     }
+
+    static var rightmostScreen: NSScreen? {
```
SenpaiHunters:
```swift
static var rightMostScreen: NSScreen? {
    NSScreen.screens.max { $0.frame.maxX < $1.frame.maxX }
}

static var leftMostScreen: NSScreen? {
    NSScreen.screens.min { $0.frame.minX < $1.frame.minX }
}
```
Slight optimisation

Loop/Managers/StashManager/StashManager.swift
```diff
@@ -0,0 +1,463 @@
...
+
+        unstashOverlappingWindows(windowToStash)
+
+        store.stashed[windowToStash.window.cgWindowID] = windowToStash
+        hideWindow(windowToStash, animate: animate)
+        startListeningMouseMoved()
+    }
+
+    /// Stop monitoring the window with the given `CGWindowID`.
+    func unstash(_ windowID: CGWindowID, resestFrame: Bool, resetFrameAnimated: Bool) {
```
SenpaiHunters: Spelling mistake here
```swift
func unstash(_ windowID: CGWindowID, resestFrame: Bool, resetFrameAnimated: Bool)
```
Should be `resetFrame`.

Loop/Managers/StashManager/StashManager.swift
```diff
@@ -0,0 +1,463 @@
...
+        mouseMoveWorkItem = workItem
+        DispatchQueue.main.asyncAfter(deadline: .now() + mouseMovedDebounceInterval, execute: workItem)
+    }
+
+    /// Handles mouse movement events to reveal or hide stashed windows.
+    /// We use the fact that `WindowEngine.windowList` returns windows sorted by z-index.
+    /// This sorting is essential because if multiple stashed windows overlap and the cursor
+    /// is over their shared area, we should only reveal the topmost window.
+    func processMouseMovement() {
+        let mouseLocation = NSEvent.mouseLocation.flipY(screen: NSScreen.screens[0])
```
SenpaiHunters, line 351:
```swift
let mouseLocation = NSEvent.mouseLocation.flipY(screen: NSScreen.screens[0])
```
I think index 0 is not always safe on multi-monitor setups. @MrKai77 will need to double-check that one, as I can't fully remember. A potential fix could be `NSScreen.main` or a screen containing the mouse location.

Loop/Window Management/Window.swift
```diff
@@ -131,19 +142,63 @@ class Window {
...
+
+        // Then set the window as main after a brief delay to ensure proper ordering
+        DispatchQueue.main.asyncAfter(deadline: .now() + 0.1) {
+            try? self.axWindow.setValue(.main, value: true)
+        }
+
+        focus()
+    }
+
+    /**
```
SenpaiHunters: Please use 
```
///
```
or
```
//
```
to denote comments instead of `/* */`.

SenpaiHunters (changes requested):

SenpaiHunters:
I also noticed the README still needs to be updated. I don't think I can find anything else. I did test it, and it all seems to be working well. We'll wait to see if Kai said anything or noticed any changes.

Loop/Managers/StashManager/StashManager.swift
guillaumecledat: The line of code was proposed by @MrKai77 ([here](https://github.com/MrKai77/Loop/pull/734#discussion_r2106281455)). My testing on a multi-monitor system did not reveal any issues. I am keeping this as unresolved until @MrKai77 provides additional input.

guillaumecledat:
Regarding the README, since it's the first thing users see, I thought you might want to write that section yourself. Also, it would be best if the example video matched the setup of your other videos (same app, same wallpaper). However, I'm happy to propose something if you'd like.

Another thing that might be missing it a custom icon for the stash action:

Screenshot 2025-06-11 at 14 53 11

SenpaiHunters:
Sure, we can do the README. Regarding an icon, I think we'll use the same source we did before. Let me look to see if we can find a nice one, and I'll let you know. If you'd like to propose anything, feel free to do so.

Loop/Managers/StashManager/StashManager.swift
mrkai77: This should be fine, we use this in other parts of the codebase as well. It works on both my single-monitor setup and multi-monitor setup.

MrKai77:
> Upon further thought, I think that my original idea of an action to show all stashed windows simultaneously could be awkward. Your solution addresses this but it does introduce some difficulty when using `cycle` with different `stash` positions.

Yes, making the keybind a toggle action would make it difficult to use in cycle actions. Personally, I think it makes sense for stash actions not to be included in cycles, since each time it's called, it would perform a different (even unexpected) operation. @Morniak, do you have a scenario where having stashes as part of a cycle is necessary? I’d be interested to hear your perspective on this, and @SenpaiHunters as well.

SenpaiHunters:
A cycle assumes idempotency in each transition—i.e., that each step in the cycle reliably produces a specific, known state. Toggling a stash that depends on the current stash position breaks this model because the result isn’t deterministic unless you track the stash state independently, which adds complexity. Stash behaviours should be treated as atomic, direct actions—not part of cycling sequences.

guillaumecledat:
I find it much more convenient to have a small number of keyboard shortcuts to remember - even if that means having to cycle through several positions - rather than having one shortcut per stash position. My usage pattern is very much "stash once and forget": I typically stash a window at the beginning of a work session and keep it there throughout. Having a cycle with multiple stash positions allows me to stash a window where I want using a single, memorable shortcut. Since I rarely stash additional windows after that, the time spent cycling through positions feels like a reasonable trade-off compared to either having to create and remember many shortcuts, or being limited to just a few stash positions.

Here’s an example of a cycle I use:
Screenshot 2025-06-12 at 13 17 05

@SenpaiHunters I believe idempotency is preserved in the example above — each cycle element represents a specific, well-defined state.

That said, this is just how I personally use the feature, and I don’t have a strong opinion on the matter. Your solution works well too (if we prevent using stash action in cycle), and I’m happy to go with it.

Last-minute thought:
Does this feature really make sense in a purely keyboard-driven workflow? A keyboard-only user could simply use the system command to hide a window (e.g. Cmd + H). Maybe we can drop the idea of a keyboard shortcut for revealing stashed windows, assuming this feature is mostly meant for mouse interaction anyway?

SenpaiHunters:
I personally don't use cycles, but I have memorized a lot of keybinds, so it's not something I tend to forget. It could be possible to add it via a cycle, but that's a bit more work. If you're able to do it and it's clean, it could be a good idea. That being said, there are benefits and trade-offs to both methods.

Regarding your last point, we want there to also be keybinds and cursor triggers, as I never touch my mouse, along with a few people who use the app. As Loop isn't made to just focus on JUST the x or y method (that being keyboard or mouse interactions), it focuses on both actions. Therefore, removing it to say "use this method" does not align with the main idea of loops, so keep it and don't discard it.

Loop/Managers/StashManager/StashManager.swift
```diff
@@ -0,0 +1,504 @@
...
+/// - Configurable animations for reveal/hide behaviors (see `Defaults[.animateStashedWindows]`).
+/// - Configurable visibility padding to determine how much of a stashed window remains visible (see `Defaults[.stashedWindowVisiblePadding]`).
+/// - Smart handling of overlapping stashed windows along the same screen edge, using vertical range tolerance.
+/// - Debounced and throttled mouse movement handling to avoid performance issues.
+/// - Automatic focus-shifting to another window when a window is hidden (optional) (see `Defaults[.shiftFocusWhenStashed]`).
+///
+/// ## Constants:
+/// - `mouseMovedDebounceInterval`: The minimum time interval (in seconds) between processing consecutive mouse move events.
+/// - `revealThrottleInterval`: The minimum time interval (in seconds) between revealing or hiding actions for a specific window.
+/// - `minimunVisibleHeightToKeepWindowStacked`:
```
mrkai77:
```suggestion
/// - `minimumVisibleHeightToKeepWindowStacked`:
```

Loop/Managers/StashManager/StashManager.swift
```diff
@@ -0,0 +1,504 @@
...
+
+// MARK: - Mouse moved listener
+
+private extension StashManager {
+    func startListeningMouseMoved() {
+        guard mouseMonitor == nil else { return }
+
+        print("StashManager: Listening for mouse moved events…")
+
+        mouseMonitor = NSEventMonitor(scope: .global, eventMask: .mouseMoved) { [weak self] _ in
```
mrkai77:
```suggestion
        mouseMonitor = NSEventMonitor(scope: .all, eventMask: .mouseMoved) { [weak self] _ in
```

This will make stashing work even when Loop is the frontmost application.

README.md
```diff
@@ -60,6 +60,12 @@ Loop can become very powerful when paired with cycles. These enable you to perfo
 
 <div><video controls src="https://github.com/user-attachments/assets/1adb1325-775d-4687-9085-71c7f775d65d" muted="false"></video></div>
 
+### Stash
+
+Hide windows at the screen edge to declutter your workspace. Hover near the edge or use a keybind to access them whenever you need.
+
+<div><video controls src="https://github.com/user-attachments/assets/8749a3bd-7a6f-4844-bb8b-e1eeacbe259c" muted="false"></video></div>
```
mrkai77: I can't seem to find the link to this video within the GitHub repo. All videos should be stored inside the `assets/graphics` folder if that helps :)

Also, thank you so much for getting around to making this, we really appreciate it!

MrKai77 (changes requested):
Tested locally for a while, works really well! All of my review comments are minor changes. Will also test this for a few days to see if I encounter any new bugs as well :)

Loop/Extensions/Defaults+Extensions.swift
```diff
@@ -53,6 +53,9 @@ extension Defaults.Keys {
     static let focusWindowOnResize = Key<Bool>("focusWindowOnResize", default: true, iCloud: true)
     static let respectStageManager = Key<Bool>("respectStageManager", default: true, iCloud: true)
     static let stageStripSize = Key<CGFloat>("stageStripSize", default: 150, iCloud: true)
+    static let animateStashedWindows = Key<Bool>("animateStashedWindows", default: true, iCloud: true)
+    static let stashedWindowVisiblePadding = Key<CGFloat>("stashedWindowVisiblePadding", default: 20, iCloud: true)
```
mrkai77, line 57: It may be helpful to re-hide windows using the new value when the user adjusts this setting.

Loop/Managers/StashManager/StashManager.swift
```diff
@@ -0,0 +1,504 @@
...
+
+        // Keep only one window as revealed
+        for revealedWindowId in store.revealed {
+            guard let revealedWindow = store.stashed[revealedWindowId] else { break }
+            hideWindow(revealedWindow, animate: animate)
+        }
+
+        let frame = window.computeRevealedFrame()
+
+        window.window.activate()
```
mrkai77: I wonder if it'd make sense to put this behind `shiftFocusWhenStashed` as well? As a window activation implies a window focus change.

MrKai77 (changes requested):

SenpaiHunters:
Looks like Kai has already outlined everything I would want to change. There is a part of me that wants to break up `processMouseMovement()`. This method is doing too much:

-  Getting mouse location  
-  Sorting windows  
-  Checking reveal state  
-  Handling reveal/hide logic  
-  Managing tolerance  

This can be broken up into small parts, using DRY (Don't Repeat Yourself), but it's not a massive thing that will have your PR denied.

SenpaiHunters (dismissed):

Loop/Managers/StashManager/StashManager.swift
guillaumecledat: Yes it make total sense 👍

README.md
guillaumecledat: I added the video to the `assets/graphics` folder and updated the README with a local path, but that doesn't seem to work. I might be missing something, but I believe that to use a URL like before, the PR needs to be merged and the video file must be present in the main Loop repository.

mrkai77: Ah yes, we can change that from our side after the PR is merged :)

MrKai77 (changes requested):
1. In `WindowAction`'s `getName()` method (line 134), it should return the window action's custom name if one is set; otherwise, it should default to "Stash."
2. In `CustomActionConfigurationView`, the "Use pixels" toggle currently appears in the position tab, but it’s only relevant for the "Unstashed Size" tab. For stash keybinds, it would be clearer to move this toggle into the "Unstashed Size" tab, directly above the sliders.
3. When creating a new Stash keybind, there’s no default selection in the position tab on first open, which can be confusing.

Also, considering the growing differences between Custom and Stash window actions, it might be worthwhile to split them into separate files to reduce direction-specific checks and logic. Let me know if you’d like me to help directly with the code (as UI-side can be a little daunting sometimes), happy to contribute to this PR too :)

image

guillaumecledat:
@MrKai77 Thanks for the feedback! I've split `CustomActionConfigurationView` into two separate views. Let me know if this works for you, or if you'd prefer that I extract shared elements into subcomponents to reduce code redundancy.

MrKai77 (approved):
Thanks for your patience! I’ve spent some time testing, and it all looks good on my end. Thank you for your support!

Issue #767, opened 2025-07-28
🐞 Settings aren't applied from copied preference plist file


injust:
### Bug Description

When I was temporarily switching to the stable version to debug another issue, I backed up the preference file at `~/Library/Preferences/com.MrKai77.Loop.plist`. However, when I restored it, it seems like none of the settings applied.

Note that I backed the preference file from the dev version and restored later to the dev version. I'm not restoring on a different version of Loop.

This was also mentioned in https://github.com/MrKai77/Loop/issues/191#issuecomment-2182341683.

### Steps To Reproduce

1. `brew install --cask loop`
2. Start Loop, update to the dev version, and configure the settings
3. Quit Loop
4. Copy `~/Library/Preferences/com.MrKai77.Loop.plist` somewhere
5. `brew uninstall --zap --cask loop`
6. `brew install --cask loop`
7. Start Loop and update to the dev version
8. Quit Loop
9. Move the preference file back to `~/Library/Preferences/com.MrKai77.Loop.plist`
10. Start Loop

### Expected Behavior

Loop starts with the settings that were previously configured.

### Actual Behavior

Loop starts with the default settings.

### macOS Version

Sequoia 15.5

### Loop Version

Version 🧪 1.3.0 (1515)

### Final Checks

- [x] My issue is written in English
- [x] My issue title is descriptive
- [x] This is a single bug (multiple bugs should be reported individually)
- [ ] I can help with fixing or developing this issue (tick if you can help).

SenpaiHunters:
This is something I should add to exporting. In honesty, currently it's only keybinds and the version key, not everything. I think it'd be easier that way. The reason it does not work is that the two versions differ, and macOS didn't "make it," if that makes sense, so it won't load the old plist settings unless the app is specified to ingest.

injust:
> This is something I should add to exporting. In honesty, currently it's only keybinds and the version key, not everything. I think it'd be easier that way.

Yes please, that would be awesome 😍

> The reason it does not work is that the two versions differ, and macOS didn't "make it," if that makes sense, so it won't load the old plist settings unless the app is specified to ingest.

I see. Seems like copying plist files around usually works, but not if the app was completely uninstalled and reinstalled.

SenpaiHunters:
@namalum If this is related to the issue, please clarify the connection. If not, kindly create a new issue.

cipher-shad0w:
[...]

PR #793, opened 2025-08-20
✨ Add settings export and import (#767)

cipher-shad0w:
issue #767

### DIFF
--- a/Loop/Extensions/Color+Extensions.swift
+++ b/Loop/Extensions/Color+Extensions.swift
@@ -117,3 +117,36 @@ extension NSColor {
         return NSColor(srgbRed: red, green: green, blue: blue, alpha: alpha)
     }
 }
+
+// MARK: - SwiftUI Color Codable support
+
+extension Color: Codable {
+    enum CodingKeys: String, CodingKey {
+        case red, green, blue, alpha
+    }
+
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        let red = try container.decode(Double.self, forKey: .red)
+        let green = try container.decode(Double.self, forKey: .green)
+        let blue = try container.decode(Double.self, forKey: .blue)
+        let alpha = try container.decode(Double.self, forKey: .alpha)
+
+        self.init(red: red, green: green, blue: blue, opacity: alpha)
+    }
+
+    public func encode(to encoder: Encoder) throws {
+        var container = encoder.container(keyedBy: CodingKeys.self)
+
+        // Convert SwiftUI Color to NSColor for component extraction
+        let nsColor = NSColor(self)
+        guard let rgbColor = nsColor.usingColorSpace(.sRGB) else {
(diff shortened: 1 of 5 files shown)


MrKai77:
I see that you added an extension for Color to conform to Codable. Take a look at: https://github.com/sindresorhus/Defaults/blob/d88a5c8cdb498e7536ef5ff88e7120e9b5874806/Sources/Defaults/Defaults%2BBridge.swift#L370

I wonder if it's possible to rely on the Defaults package for the serialization of certain values, also since Loop's saved plist uses that format too?

cipher-shad0w:
I will take a look at that

cipher-shad0w:
I investigated the suggested ColorBridge approach and concluded that it's not suitable for our JSON export/import feature. Here's a detailed explanation:

Defaults.Serializable ≠ Codable:
- Defaults.Serializable is optimized for UserDefaults (Property Lists)
- Codable is designed for JSON serialization with Encoder/Decoder pattern
- These are two completely different serialization paradigms

The ColorBridge from Defaults serializes Color as:
```swift
public typealias Serializable = Any

// Serializes to:
[colorSpace, components] as [Any]
```

Problem: Any is not Codable-conformant, so I can't use it in JSON structures.

If I tried:
```swift
extension Color: Defaults.Serializable {
    public static let bridge = Defaults.ColorBridge()
}

struct SavedSettingsFormat: Codable {
    let customAccentColor: Color?  // ❌ Error: Color is not Codable
}
```

I get: `error: cannot automatically synthesize 'Decodable' because 'Color?' does not conform to 'Decodable'`

The current ColorData implementation is optimal for our use case:

1. JSON-compatible: Serializes to simple numbers
2. Lossless: Preserves RGBA components exactly
3. Platform-agnostic: Works everywhere JSON works
4. Separation of concerns:
   - UserDefaults continues using Defaults' built-in Color support
   - JSON export/import uses ColorData

cipher-shad0w:
In short, I can simplify the Color class, but it doesn't work when I try to use the default pack.

MrKai77:
That means that Loop would be reading/writing color data in two completely different formats, I don't think this is ideal. 

I also believe that it's very not clean to be manually transcoding the user's entire settings into JSON directly from UserDefaults keys, it just seems a little inefficient when we know that we want to export all keys.

I think it's worth revisiting the format in which we store exported settings, as well as how to make this as simple as possible. Would it be more suitable to export a copy of the stored UserDefaults plist?

MrKai77:
Also cc'ing @SenpaiHunters here :)

cipher-shad0w:
I will take a closer look at this over the next few days and find a better solution for exporting and importing the settings. I hadn't thought of the plist file, which is of course the easiest way and means you don't have to save everything individually. 
The only limitation is that you then export all settings and can no longer select which ones to export. I will implement this with the plist. 
Thanks for the detailed feedback. @MrKai77

cipher-shad0w:
I'll make a new PR for that

Issue #754, opened 2025-06-14
✨ Better Keybind Workflow


LeBeaBae:
### Feature Request Description

Organizing menus like the list of options for keybindings is daunting and very subjective, but I think that a couple of rules would be helpful. It would be smart to have the "Custom" and "Cycle" options at the top of the list; They are unique features and offer a very interesting concept. It would be smart to *advertise* this feature, and it should be at the top. You could also replace the bottom of the menu with my idea from #753 .

### Screenshots

![Image](https://github.com/user-attachments/assets/8e239a12-f2be-4cf5-8406-4800319685d3)
The options to move!

### Final Checks

- [x] My issue is written in English
- [x] My issue title is descriptive
- [x] This is a single feature request (multiple feature requests should be reported individually)
- [x] I can help with the improvement or development of this feature (tick if you can help)

MrKai77:
See https://github.com/MrKai77/Loop/pull/794#issuecomment-3210567849

PR #794, opened 2025-08-20
✨ Move the "More" keybinds to the top

cipher-shad0w:
Issue #754

### DIFF
--- a/Loop/Luminare/Settings/Keybindings/KeybindItemView.swift
+++ b/Loop/Luminare/Settings/Keybindings/KeybindItemView.swift
@@ -49,9 +49,9 @@ struct KeybindItemView: View {
 
     var moreSection: PickerSection<WindowDirection> {
         if cycleIndex != nil { // If this is a cycling keybind
-            .init(.init(localized: "More"), [WindowDirection.custom])
+            .init(.init(localized: "Special"), [WindowDirection.custom])
         } else {
-            .init(.init(localized: "More"), [WindowDirection.custom, WindowDirection.cycle])
+            .init(.init(localized: "Special"), [WindowDirection.custom, WindowDirection.cycle])
         }
     }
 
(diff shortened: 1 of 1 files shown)


MrKai77:
Hmm, if we place this item at the top, I don't think that "More" is a good header for it. Maybe "Customizable" or similar?

cipher-shad0w:
I have renamed the header from 'More' to 'Special'. Is that ok, or should I use a different name?

Loop/Luminare/Settings/Keybindings/KeybindItemView.swift
```diff
@@ -49,9 +49,9 @@ struct KeybindItemView: View {
 
     var moreSection: PickerSection<WindowDirection> {
         if cycleIndex != nil { // If this is a cycling keybind
-            .init(.init(localized: "More"), [WindowDirection.custom])
```
SenpaiHunters: Change this back to "More" along with the other options.

SenpaiHunters (changes requested):

SenpaiHunters (approved):

SenpaiHunters:
Going to cc @MrKai77 on this for the review, as it's up to him if he wishes to change the order so he'll have the final say.

MrKai77:
"More" only makes sense if it's at the bottom of the whole list. "Special" isn't really descriptive of what the actions do, which isn't ideal if it's the first thing the user sees. Something like "configurable" or "flexible" makes more sense, since that's the main distinctive difference.

MrKai77:
Actually, after discussing this a little more with the team, we have decided not to move forward with this request, since both the "custom" and "cycle" actions are meant to be power user features. Average users won't need to use them, on which case moving the items to the top simply adds more complexity.

cipher-shad0w:
Okay

PR #796, opened 2025-08-26
Add settings export and import functionality

### LINKED ISSUE #767: 🐞 Settings aren't applied from copied preference plist file

cipher-shad0w:
This update introduces the ability to export and import settings in Loop, addressing the issue where settings from a copied preference plist file were not applied. The changes include new notifications for successful import and export of settings, as well as additional alert messages for various scenarios related to settings management.

Fixes #767


MrKai77:
Hey @cipher-shad0w, thanks a lot for the PR! After reviewing it, we’ve decided not to move forward with this approach. While the idea makes sense, there are a few concerns; for example, the UserDefaults plist often contains more than just Loop’s preferences, such as iCloud sync data and other macOS information. Because of this, I think it’d be better to take a step back and rethink how this should work before implementing it.

I also noticed that RectangleTranslationLayer was removed. I’m not entirely sure why, but it seems like something we should keep around. Could you share your reasoning for deleting it?

cipher-shad0w:
I removed RectangleTranslationLayer because I thought it would become redundant with the new general settings import/export system. My idea was that if users can now export and import all their Loop settings, they wouldn't need this Rectangle import functionality anymore and I'm working on another import for Rectangle and other apps

I have considered that if only certain settings are to be imported, you can import specific settings and compare them with the existing ones. This can be easily adjusted so that you do not have to import and export the entire plist file. How about if I adjust the import and export function again and then perform a targeted import and export?

MrKai77:
Thanks for clarifying! We really appreciate the effort you’ve put into this. That said, we don’t think this implementation is the right fit for Loop. We do want to add an import/export system in the future, but it’ll need to be approached differently and with a cleaner design before we can move forward. There are also several complexities in the current approach that we aren’t entirely satisfied with. For example, plist files aren’t as easy to work with in a raw text editor compared to other formats. Simplification is especially important here, since all future versions of Loop will need to maintain backwards compatibility with this format. If we choose something that needs to change later, the legacy support code inside Loop will only keep growing.

PR #800, opened 2025-08-30
📝  Standardize date format in file headers

cipher-shad0w:
Update file headers to use a consistent date format (YYYY-MM-DD) for improved readability and standardization.


MrKai77:
I don’t think this change is necessary, as it doesn’t improve really readability and may even reduce some of the character that makes this project unique ;)
Personally, I find the current version a fun and subtle reminder that Loop was created by people from all over the world.

cipher-shad0w:
I understand, that's a good reason.
I was just too much of a perfectionist :)

PR #814, opened 2025-09-07
✨ Introduce AccentColorController

MrKai77:
This PR introduces a new class, `AccentColorController`, which acts as a singleton to manage the app’s accent color state globally.

Compared to the previous implementation, this approach has several advantages:
- When the accent color mode is set to wallpaper, Loop will no longer overwrite a user’s custom accent color preferences
- Because it conforms to `ObservableObject`, any views that depend on this class will automatically update when the accent color changes

The second point is the most impactful change. Previously, many views such as the radial menu and preview window relied on `.onChange` callbacks to refresh their accent color. With this new design, those updates happen automatically. Loop’s settings window also benefits from being more responsive to accent color updates.

This PR also adds a new `accentColorMode` key to Defaults. It replaces the older `useSystemAccentColor` and `processWallpaper` keys with a single enum of type `AccentColorMode`. This simplifies how Loop handles accent colors, but also required a way to migrate user settings.

To handle that migration, this PR introduces a `DataPatcher` enum. Its `run()` function is executed at startup and tracks which patches have already been applied. It automatically transitions old user defaults to the new format, and it provides a foundation for handling future data migrations as well.


SenpaiHunters (approved):
I've been testing it locally for the past two days with all three color modes. I can't see any difference, which is good, so I'll accept it and merge it!

PR #820, opened 2025-09-14
✨ Add directional screen switching

### LINKED ISSUE #817: ✨ Add directional screen switching

Currently, it is only possible to assign shortcuts for “previous screen” and “next screen” to move a window across monitors.
This works, but becomes unintuitive when working with setups that are not in a simple horizontal row (e.g., one monitor above another).

 Additional shortcut options for moving a window to:
	•	Left screen
	•	Right screen
	•	Screen above
	•	Screen below

This would allow more precise and natural navigation in multi-monitor setups.

### Screenshots

_No response_

### Additional Context

This feature would especially benefit users with vertical or 2x2 monitor arrangements. It complements the existing “previous/next screen” shortcuts without replacing them.

### Final Checks

- [x] I can help with the improvement or development of this feature (tick if you can help)

cipher-shad0w:
#817 

Screenshot 2025-09-14 at 14 15 23


Loop/Utilities/ScreenUtility.swift
```diff
@@ -8,6 +8,29 @@
 import SwiftUI
 
 enum ScreenUtility {
+    // MARK: - Screen Cache
+
+    private static var cachedScreens: [NSScreen]?
+    private static var cacheTimestamp: Date?
+    private static var cachedScreenCount: Int = 0
+    private static let cacheValidityDuration: TimeInterval = 0.5
+    private static let cacheLock = NSLock()
```
mrkai77: Rather than relying on an NSLock, could we encapsulate the variables requiring concurrency safety within an actor? Our long-term objective for Loop is to fully adopt modern Swift concurrency APIs.

Loop/Utilities/ScreenUtility.swift
```diff
@@ -108,4 +306,88 @@ private extension Array where Element: Hashable {
...
         return nil
     }
+
+    private func directionalScreen(
+        from item: Element,
+        isCandidate: (NSScreen, NSScreen, CGFloat) -> Bool,
+        overlap: (NSScreen, NSScreen) -> CGFloat,
+        distance: (NSScreen, NSScreen) -> CGFloat
+    ) -> Element? {
+        guard let screen = item as? NSScreen else { return nil }
```
mrkai77: The properties `directionalScreen`, `left`, `right`, `top`, and `bottom` should not be included in an Array extension. This extension is meant for arrays whose elements conform to `Hashable`, not specifically for arrays of `NSScreen`. Instead, these should be added to the `ScreenUtility` enum, and the methods should remain private.

This is particularly relevant to this line where `item` is being optionally cast to `NSScreen`, a cast that wouldn’t be necessary if these properties weren’t placed in this extension.

Loop/Utilities/ScreenUtility.swift
```diff
@@ -29,8 +52,159 @@ enum ScreenUtility {
...
+            return bottomScreen
+        }
+
+        guard canRestartCycle else {
+            return nil
+        }
+
+        let currentFrame = screen.frame
+
+        let overlappingScreens = screens.filter { otherScreen in
```
mrkai77: I see this `overlappingScreens` computation in all of the `leftScreen`, `rightScreen`, `topScreen` and `bottomScreen` functions. Can we extract this into a separate function to reduce duplication?

Loop/Utilities/ScreenUtility.swift
```diff
@@ -29,8 +52,159 @@ enum ScreenUtility {
...
+        defer { cacheLock.unlock() }
+
+        cachedScreens = nil
+        cacheTimestamp = nil
+        cachedScreenCount = 0
+    }
+
     // MARK: Private
 
+    private static func leftmostScreen(in screens: [NSScreen]) -> NSScreen? {
```
mrkai77: The `leftmostScreen`, `rightmostScreen`, `topmostScreen` and `bottommostScreen` aren't needed as individual functions; they're only one-liners, and are each only called from a single part of the codebase. Instead, directly place these functions within the functions that need these properties, and rename `fallback` to `[direction]mostScreen` accordingly.

Loop/Utilities/ScreenUtility.swift
```diff
@@ -29,8 +52,159 @@ enum ScreenUtility {
...
+            return topmostOverlapping
+        }
+
+        let fallback = topmostScreen(in: screens)
+        return fallback
+    }
+
+    // MARK: - Cache Management
+
+    static func invalidateScreenCache() {
```
mrkai77: If this isn't needed outside of the enum, make it private.

MrKai77 (changes requested):

Loop/Utilities/ScreenUtility.swift
mrkai77: It looks like you replaced it with a `DispatchQueue(label: "com.loop.screenUtility.cache", attributes: .concurrent)`. However, since the `.concurrent` attribute is used, this may actually undermine the goal of making the property thread-safe.

Would it be possible to use an `actor` here instead? Do realize this isn’t a blocking change - switching to an actor would require adopting async/await throughout, which could complicate this PR. Still, I’d be interested to hear your thoughts on this approach!

cipher-shad0w: Ah, I think I understand you now. Do you mean something like this:

@MainActor
final class ScreenCache {
    ...
}

mrkai77: Yes, that would work, and for this PR's purposes, might actually recommend that. As in, adding `@MainActor` to the existing ScreenUtility enum. Doing what I show below may be overkill for this PR :)

---

What I originally meant was something like:

```swift
fileprivate actor ScreenCache {
    // thread-safe properties here
}

enum ScreenUtility {
    // ScreenCache instance in here, and all methods call it to access thread-safe properties
}
```

---

But truthfully, I also don't think thread safety is *too* important here. If this requires too many changes in other files, it is also fine to not have any actor assignments here, and simply rely on the static cache. We can address this in a future PR if needed as well.

PR #842, opened 2025-10-02
✨ Option to suppress Mission Control when window snapping

### LINKED ISSUE #838: ✨ Window snapping compatibility with drag-to-top to enter Mission Control

I use Loop's window snapping, however it is not compatible with these macOS features (System Settings > Desktop & Dock):

- Drag windows to menu bar to fill screen
- Drag windows to top of screen to enter Mission Control

I particularly miss the 2nd one, because it's the best way I've found to move windows between spaces (at least until https://github.com/MrKai77/Loop/issues/223).

In Rectangle, it's possible to disable the top snap area to maintain compatibility with these features. This would be a good stopgap compatibility fix.

The ultimate goal would be to have Loop be compatible with "Drag windows to top of screen to enter Mission Control", so dragging windows to the menu bar would still trigger Loop's preview animation.

### Screenshots

_No response_

### Additional Context

I thought I opened a feature request for this a while ago, but I couldn't find it. I think this is the last item needed to match all the Rectangle functionality I used!

### Final Checks

cipher-shad0w:
Adds a new setting to disable the top snap area (menu bar) in window snapping, allowing macOS native features like 'Drag windows to top of screen to enter Mission Control' to work.

Changes:
- Add 'disableTopSnapArea' setting in Defaults+Extensions
- Update WindowDirection+Snapping to skip top snap when disabled
- Add UI toggle in BehaviorConfiguration under Window section
- Update cursor warp behavior to respect the new setting

ISSUE: #838


Loop/Core/WindowDragManager.swift
```diff
@@ -39,11 +39,14 @@ class WindowDragManager {
...
-                    if let frame = NSScreen.main?.displayBounds,
-                       let mouseLocation = CGEvent.mouseLocation {
-                        if mouseLocation.y == frame.minY {
-                            let newOrigin = CGPoint(x: mouseLocation.x, y: frame.minY + 1)
-                            CGWarpMouseCursorPosition(newOrigin)
+                    // Only warp cursor away from top edge if top snap area is enabled
+                    if !Defaults[.disableTopSnapArea] {
+                        if let frame = NSScreen.main?.displayBounds,
+                           let mouseLocation = CGEvent.mouseLocation {
+                            if mouseLocation.y == frame.minY {
```
mrkai77: Can we consolidate the `Defaults[.disableTopSnapArea]`, display bounds calculation and the line tagged on this comment’s `if` statements all inside a single `if` statement instead?

Loop/Extensions/Defaults+Extensions.swift
```diff
@@ -44,6 +44,7 @@ extension Defaults.Keys {
     static let hideMenuBarIcon = Key<Bool>("hideMenuBarIcon", default: false, iCloud: true)
     static let animationConfiguration = Key<AnimationConfiguration>("animationConfiguration", default: .fast, iCloud: true)
     static let windowSnapping = Key<Bool>("windowSnapping", default: false, iCloud: true)
+    static let disableTopSnapArea = Key<Bool>("disableTopSnapArea", default: false, iCloud: true)
```
mrkai77: Better wording for this might instead be `suppressMissionControlOnTopDrag`, and have it on by default.

Loop/Window Management/Window Action/WindowDirection+Snapping.swift
```diff
@@ -20,7 +21,7 @@ extension WindowDirection {
             newDirection = WindowDirection.processLeftSnap(mouseLocation, screenFrame)
         } else if mouseLocation.x > ignoredFrame.maxX {
             newDirection = WindowDirection.processRightSnap(mouseLocation, screenFrame)
-        } else if mouseLocation.y < ignoredFrame.minY {
+        } else if mouseLocation.y < ignoredFrame.minY, !Defaults[.disableTopSnapArea] {
```
mrkai77: I think it makes sense to only disable the blocking of Mission Control, rather than fully blocking the top snap area. This should work fine, as this was how Loop used to work in earlier versions. Otherwise, users would be missing the maximize snap. Open to input here though :)

MrKai77 (changes requested):

Loop/Extensions/Defaults+Extensions.swift
mrkai77: I noticed that this comment was marked as resolved, but my request was for this to be enabled by default. It appears that it’s still disabled by default; could you take another look?

Loop/Settings Window/Settings/Behavior/BehaviorConfiguration.swift
```diff
@@ -80,6 +81,17 @@ struct BehaviorConfigurationView: View {
                 LuminareToggle("Window snapping", isOn: $windowSnapping)
             }
 
+            if windowSnapping {
+                LuminareToggle(isOn: $suppressMissionControlOnTopDrag) {
+                    Text("Allow Mission Control on top drag")
```
mrkai77: `suppressMissionControlOnTopDrag` and `Allow Mission Control on top drag` are opposites, reword this text :)

Loop/Settings Window/Settings/Behavior/BehaviorConfiguration.swift
```diff
@@ -80,6 +81,17 @@ struct BehaviorConfigurationView: View {
...
+                        .padding(.trailing, 4)
+                        .luminarePopover(attachedTo: .topTrailing) {
+                            Text("Allows mission control to work while still keeping top snap functionality.")
```
mrkai77:
```suggestion
                            Text("Whether to allow mission control to open when windows are dragged to the top.")
```

MrKai77 (changes requested):

Loop/Extensions/Defaults+Extensions.swift
cipher-shad0w: Of course, it just missed

Loop/Window Management/Window Action/WindowDirection+Snapping.swift
mrkai77: What's the reason for this addition?

cipher-shad0w: That was from another commit, so I still had `!Defaults[.disableTopSnapArea]` in there. I simply overlooked the fact that I had to remove it again

PR #843, opened 2025-10-02
🐞 Adjust size increment slider step value in AdvancedConfigurationView

### LINKED ISSUE #836: 🐞 Size Increment number field forces first digit up to minimum value of 5
### Bug Description

Size Increment defaults to 20px. I changed it while experimenting with the settings and it was difficult to set it back to 20px.

The slider changes values in increments of 5px, so the closest I could get it was 18px. This isn't a huge deal though, because I expected to be able to set an exact value using the number field.

However, when using number field, typing `2` sets it to the minimum value of 5px, and then typing `0` sets it to 50px.

I eventually gave up and unset it using `defaults`.

While writing up this issue, I figured out how to do it properly: After the `2` gets turned into a `5`, backspace the `5` and type `2` again, which will actually input a `2`.

It's nice that it's actually possible, but maybe there's a better approach so that typing `20` actually gets you the correct value the first time.

### Steps To Reproduce

1. Change the Size Increment setting
2. Try to set it back to the default of 20px

### Expected Behavior

.

### Actual Behavior

.

### Screenshots

_No response_

### macOS Version

Sequoia 15.7

### Loop Version

Version 🧪 1.4.0 (1546)

### Additional Context

_No response_

### Final Checks

cipher-shad0w:
## Problem

The Size Increment setting had a critical usability issue where typing values in the number field would behave unexpectedly:

- Typing `2` would immediately jump to `5` (the minimum value)
- Then typing `0` would result in `50` instead of the intended `20`
- This made it extremely difficult to set the default value of `20px`

As described in issue #836, the only workaround was to:
1. Type `2` (which becomes `5`)
2. Backspace to delete the `5`
3. Type `2` again (which finally works)
4. Type `0` to get `20`

## Root Cause

The `LuminareSlider` for Size Increment was configured with a `step` parameter of `4.5`, which created the following valid values:

5, 10, 15, 20, 25, 30, 35, 40, 45, 50

Now:
- The default value of `20px` is included in the valid range
- Typing `20` in the number field works correctly on the first try
- All increment values are whole numbers

## Changes

- `Loop/Settings Window/Loop/AdvancedConfiguration.swift`: Changed `step: 4.5` to `step: 5`

Closes #836

### DIFF
--- a/Loop/Settings Window/Loop/AdvancedConfiguration.swift
+++ b/Loop/Settings Window/Loop/AdvancedConfiguration.swift
@@ -202,7 +202,7 @@ struct AdvancedConfigurationView: View {
                 "Size increment",
                 value: $sizeIncrement.doubleBinding,
                 in: 5...50,
-                step: 4.5,
+                step: 5,
                 format: .number.precision(.fractionLength(0...0)),
                 clampsUpper: false,
                 suffix: Text("px")


MrKai77:
This seems more like a workaround rather than a fix. It’s possible this bug affects other parts of the app too. It would make sense to modify Luminare instead, and add the formatting/clamping of the number after the user submits their value, rather than applying it on every mutation of the inputted variable (if that makes sense). Let me know if you’d like more information about that!

cipher-shad0w:
Makes sense, I just thought you could fix it that way for now and I didn’t see the point of the 4.5 why it was used. And with 5 the bug no longer occurs. Let me know if I can help you in any way with this issue

MrKai77:
Ah, I can see why that would feel random: it's set to 4.5 so that there are 10 equal increments within the range of `5...50` (so `45/10`).
If you'd like to try and fix this issue, you are free to check out Luminare and open a PR over there ;)

PR #844, opened 2025-10-02
🐞 Fix screen handling in WindowDragManager and PreviewController for accurate mouse position tracking

### LINKED ISSUE #835: 🐞 window jump when dragging on the secondary (external) display
### Bug Description

I have seen a weird behavior when I drag app windows when I am connected to a second display.
When I drag a window on the external display(secondary display) around (drag it within the screen, not to the main display), the app jumps to the bottom of the main screen (mac display), I am not sure if it's a bug related to Loop or macOS 15.7. Since Loop is the only window manager app I have on this mac, here I am. Noticed this recently when I work on the second display, multiple times a day.

### Steps To Reproduce

- Connect mac to a second display, mac screen set to main screen, 
- drag a window on the secondary display

### Expected Behavior

The app should not jump to the other display.

### Actual Behavior

App window jumps.

### Screenshots

_No response_

### macOS Version

15.7

### Loop Version

v1.3.0

### Additional Context

_No response_

### Final Checks

cipher-shad0w:
## Problem

- Windows would unexpectedly jump from the external display to the main display
- The jump occurred while dragging windows within the same secondary screen (not between displays)
- Window snapping and preview indicators would appear on the wrong screen

## Root Cause
The bug was caused by hardcoded references to NSScreen.screens[0] in coordinate conversion code. In macOS, screens[0] always refers to the primary display, regardless of where the mouse or window is located.

The problem occurred in the flipY() coordinate conversion function, which translates between:

- macOS coordinate system: Origin (0,0) at bottom-left, Y increases upward
- AppKit coordinate system: Origin (0,0) at top-left, Y increases downward
When dragging on a secondary display, the code was using the main screen's coordinate space for conversions instead of the actual screen's coordinate space, causing incorrect frame calculations that resulted in windows jumping to the main display.

## Solution
Replaced all hardcoded NSScreen.screens[0] references with the appropriate screen context:

- Use NSScreen.screenWithMouse to get the screen where the mouse is currently located
- Use the already-determined screen parameter instead of assuming the main screen
- Pass the correct screen parameter through function calls
This ensures coordinate conversions always use the correct screen's frame as reference.

## Expected Behavior After Fix
- Windows stay on the secondary display when dragging
- Window snapping works correctly on all displays
- Preview indicators appear on the correct screen
- Coordinate calculations are accurate across all display configurations
- Loop's settings window can be properly positioned on any display

Closes #835

### DIFF
--- a/Loop/Core/WindowDragManager.swift
+++ b/Loop/Core/WindowDragManager.swift
@@ -74,7 +74,11 @@ class WindowDragManager {
     }
 
     private func setCurrentDraggingWindow() {
-        let mousePosition = NSEvent.mouseLocation.flipY(screen: NSScreen.screens[0])
+        guard let currentScreen = NSScreen.screenWithMouse else {
+            return
+        }
+        
+        let mousePosition = NSEvent.mouseLocation.flipY(screen: currentScreen)
 
         do {
             guard
@@ -136,9 +140,8 @@ class WindowDragManager {
             return
         }
 
-        let mainScreen = NSScreen.screens[0]
-        let mousePosition = NSEvent.mouseLocation.flipY(screen: mainScreen)
-        let screenFrame = screen.frame.flipY(screen: mainScreen)
+        let mousePosition = NSEvent.mouseLocation.flipY(screen: screen)
+        let screenFrame = screen.frame.flipY(screen: screen)
 
         previewController.setScreen(to: screen)
 
--- a/Loop/Window Action Indicators/Preview Window/PreviewController.swift
+++ b/Loop/Window Action Indicators/Preview Window/PreviewController.swift
@@ -103,7 +103,7 @@ final class PreviewController {
             screen: screen,
             isPreview: true
         )
-        .flipY(maxY: NSScreen.screens[0].frame.maxY)
+        .flipY(maxY: screen.frame.maxY)
 
         // What is the screen's frame
         print("Target frame: \(targetWindowFrame)")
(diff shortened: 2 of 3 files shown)


Loop/Core/WindowDragManager.swift
```diff
@@ -74,7 +74,11 @@ class WindowDragManager {
     }
 
     private func setCurrentDraggingWindow() {
-        let mousePosition = NSEvent.mouseLocation.flipY(screen: NSScreen.screens[0])
```
mrkai77, line 77: The reason we flip Y on the first screen of the `NSScreen.screens` array is because the first display has the `(0, 0)` origin point. All other displays are placed in their exact coordinate point (for example, screen 2 might be at `(0, 1080)`. So flipping the Y based on the screen that contains the mouse is unfortunately incorrect.

MrKai77 (commented):
Unfortunately, the solution here is not correct, as much as I would have also liked the solution to be this simple :(
Out of curiosity, were you able to reproduce this bug on your end? Flipping the Y based on a non-first screen should evidently lead to issues.

cipher-shad0w:
> Unfortunately, the solution here is not correct, as much as I would have also liked the solution to be this simple :( Out of curiosity, were you able to reproduce this bug on your end? Flipping the Y based on a non-first screen should evidently lead to issues.


Yeah, I had the same issue in the last few days, but it appeared randomly and I thought it was a failure on my side. In the end, I couldn't reproduce the bug with my implementation. I had actually only tried something, and since then, the bug hasn’t occurred again. It would be too good to be true if the bug were so easy to solve. I would now work with the bug and then try to solve it, or at least consult the developer who implemented this, because I want to learn how you would implement it.

MrKai77:
> I would now work with the bug and then try to solve it, or at least consult the developer who implemented this, because I want to learn how you would implement it.

Thanks for the effort though! I am the one who implemented this, so you are free to ask me questions either here or over on the Discord server if you have any questions :)

cipher-shad0w:
Thx man. I’ve really learnt a lot from you. And thank you for helping me to get better at swift.

Ender-Wang:
I tried v1.2 and no such glitch, maybe worth comparing the two versions, clue may be hidden somewhere there. @cipher-shad0w  @MrKai77

PR #861, opened 2025-10-28
✨ Add proportional window sizing for screen transitions

### LINKED ISSUE #713: ✨ Moving to next screen should keep window size

When moving a window between screens of different sizes, the current behavior can be inconsistent. If the initial screen is smaller, the window retains its original size. However, when moving to a larger screen, users typically prefer the window to remain maximized if it was maximized on the first screen. To improve this, the window resizing mechanism should calculate size based on screen percentage rather than absolute pixel dimensions.

### Screenshots

_No response_

### Additional Context

_No response_

### Final Checks

cipher-shad0w:
Closes #713

   Implements proportional window sizing when moving windows between displays of different sizes. Previously, windows would retain their absolute pixel dimensions
   when transitioning between screens, leading to inconsistent behavior. This PR introduces screen-aware calculations that maintain window proportions relative to
   each display's dimensions.

### DIFF
--- a/Loop/Window Management/Window Action/WindowAction.swift
+++ b/Loop/Window Management/Window Action/WindowAction.swift
@@ -423,7 +423,7 @@ extension WindowAction {
     ///   - bounds: The new screen bounds to apply the proportions to
     /// - Returns: A new `CGRect` with the proportional frame applied to the new bounds
     private func applyProportionalFrame(_ proportionalFrame: CGRect, _ bounds: CGRect) -> CGRect {
-        CGRect(
+        return CGRect(
             x: bounds.origin.x + (bounds.width * proportionalFrame.minX),
             y: bounds.origin.y + (bounds.height * proportionalFrame.minY),
             width: bounds.width * proportionalFrame.width,
--- a/Loop/Window Management/Window Action/WindowAction.swift
+++ b/Loop/Window Management/Window Action/WindowAction.swift
@@ -222,8 +222,9 @@ struct WindowAction: Codable, Identifiable, Hashable, Equatable, Defaults.Serial
     ///   - disablePadding: whether to disable padding. `true` when calculating non-AX-usage frames, such as for angle calculations in radial menu or in config UI.
     ///   - screen: the screen on which the bounds are located. Only used to determine if padding should be applied (see `getBounds()`).
     ///   - isPreview: ensures that when manipulating the preview window, the last target frame does not affect the actual resizing of the window.
+    ///   - proportionalFrame: optional proportional frame when moving between screens. Values should be between 0.0 and 1.0.
     /// - Returns: the calculated frame for the specified window action.
-    func getFrame(window: Window?, bounds: CGRect, disablePadding: Bool = false, screen: NSScreen? = nil, isPreview: Bool = false) -> CGRect {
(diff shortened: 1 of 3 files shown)


Loop/Window Management/Window Manipulation/WindowTransformAnimation.swift
```diff
@@ -86,7 +86,7 @@ final class WindowTransformAnimation: NSAnimation {
                 window.size = newFrame.size
             }
 
-            lastWindowFrame = window.frame
+            lastWindowFrame = newFrame
```
MrKai77: Can you explain this change?

Loop/Window Management/Window Action/WindowAction.swift
```diff
@@ -293,12 +294,18 @@ extension WindowAction {
...
     /// - Returns: the calculated target frame for the specified window action.
-    private func calculateTargetFrame(_ direction: WindowDirection, _ window: Window?, _ bounds: CGRect, _ isPreview: Bool) -> CGRect {
+    private func calculateTargetFrame(_ direction: WindowDirection, _ window: Window?, _ bounds: CGRect, _ isPreview: Bool, proportionalFrame: CGRect? = nil) -> CGRect {
         var result: CGRect = .zero
 
         if direction.frameMultiplyValues != nil {
-            result = applyFrameMultiplyValues(bounds)
+            // When moving between screens with a proportional frame, use proportional sizing
+            if let proportionalFrame {
+                result = applyProportionalFrame(proportionalFrame, bounds)
```
MrKai77: I fail to understand why `applyProportionalFrame` only if `direction.frameMultiplyValues != nil` and `proportionalFrame` is also not `nil`. Does it make sense to keep it inside this if statement, or is there a more specific, verbose way we can move this?

Since, from my understanding, we only use `proportionalFrame` when changing screens, why not check for `direction.willChangeScreen` instead?

cipher-shad0w:
This implementation applies proportional sizing only to standard window positions (halves, quarters,
  thirds, fourths, maximize, etc.) when moving between screens.

  What's supported ✅

  - All standard positions with frameMultiplyValues (halves, quarters, thirds, fourths)
  - Maximize and almost-maximize actions

  What's NOT supported ❌

  - Custom Actions - These will use their defined pixel/percentage values as-is
  - Special actions (center, undo, move, grow/shrink, etc.)

  Technical reasoning

  The feature checks for action.direction.frameMultiplyValues != nil to determine if proportional sizing
  should be applied. Custom actions don't have frameMultiplyValues as they define their dimensions through
   different properties (width, height, unit, positionMode, etc.).

MrKai77:
This is a really good attempt at tackling the issue, thank you so much for taking the time to dive into it! Unfortunately, the approach isn’t quite correct. The main sign that this method isn’t ideal is that the preview window would no longer accurately reflect the frame of the window on the new screen. Action frames shouldn’t be calculated inside WindowEngine; its role should focus on *executing* them instead.

The correct approach would be to calculate the proportional frame within `LoopManager.changeAction`, right when the user selects the action. This both produces the right behavior and makes the overall implementation much simpler.

While reviewing this PR, I ended up putting together my own solution for the issue, which I’ll be merging (#874). I’d encourage you to take a look at it, I’d be more than happy to walk you through any of the changes if you’d like! Once again, I really appreciate your effort and initiative here. It means a lot that you took the time to work on this :)

PR #866, opened 2025-10-29
✨ Add support for side-independent trigger keys

### LINKED ISSUE #823: ✨ Allow Alt+Left and Alt+Right to be used together as the same trigger

Make the keyboard trigger recognize both Alt+Left and Alt+Right as the same logical trigger (instead of only one side), so users can activate the action using either Alt key.

Why ?

Many users instinctively use whichever Alt key is closest to the hand they’re using at the moment (left or right). Current behavior only accepts a single Alt-side for a feature, which is inconsistent and reduces accessibility and ergonomics. Supporting both Alt keys for the same shortcut improves discoverability and user comfort without changing existing keyboard layouts.

### Screenshots

_No response_

### Additional Context

_No response_

### Final Checks

MrKai77:
This toggle lets users disable Loop’s default behavior of distinguishing between left and right modifier keys.

For example, if the trigger key is left command, turning off “Treat left and right keys differently” will make Loop respond to both the left and right command keys.

### DIFF
--- a/Loop/Core/Triggers/KeybindObserver.swift
+++ b/Loop/Core/Triggers/KeybindObserver.swift
@@ -30,6 +30,11 @@ final class KeybindObserver {
 
     private var useTriggerDelay: Bool { Defaults[.triggerDelay] > 0.1 }
     private var doubleClickToTrigger: Bool { Defaults[.doubleClickToTrigger] }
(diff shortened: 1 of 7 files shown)


SenpaiHunters (commented):
This fixes the main issue I spoke about.

SenpaiHunters (approved):

PR #870, opened 2025-10-30
✨ Add directional focus navigation for windows

### LINKED ISSUE #824: ✨ Shortcut for focus change

Currently there is no built-in way to quickly move the focus between windows. For example, when working in a split-screen layout, if the focus is on the left window, the user must manually click the right window to continue working.

It would be useful to have a new action that  move focus explicitly in a direction (focus left, focus right, focus up, focus down) (this should be looped).

### Final Checks

- [x] I can help with the improvement or development of this feature (tick if you can help)

cipher-shad0w:
Closes #824

Implements functionality to focus on windows in specified directions (up, down, left, right) and enhance the window management experience.


Loop/Utilities/ScreenUtility.swift
```diff
@@ -234,6 +234,145 @@ enum ScreenUtility {
     }
 }
 
+// MARK: - Generic Directional Navigation
+
+extension ScreenUtility {
```
mrkai77: Why are these functions inside ScreenUtility? It seems like ScreenUtility itself doesn't call any of these functions.

It would make sense to extract the existing logic inside ScreenUtility for directional screen switching, then move that into its own enum named something like "DirectionalNagivationUtility". Then, call this enum from both ScreenUtility and WindowUtility.

Loop/Window Management/Window/WindowUtility+FocusNavigation.swift
```diff
@@ -0,0 +1,120 @@
...
+//
+//  Created by cipher-shad0w on 2025-10-30.
+//
+
+import AppKit
+import OSLog
+import SwiftUI
+
+extension WindowUtility {
+    private static let focusLogger = Logger(category: "WindowUtility+FocusNavigation")
```
mrkai77: No need to make another logger for navigation; just call WindowUtility's logger and prepend [FocusNagivation] before the actual log string :)

Loop/Window Management/Window/WindowUtility+FocusNavigation.swift
```diff
@@ -0,0 +1,120 @@
...
+
+    /// Finds the next window to focus in the specified direction.
+    /// - Parameters:
+    ///   - currentWindow: The currently focused window to navigate from, or nil to navigate from screen center
+    ///   - direction: The direction to search for the next window (focusUp, focusDown, focusLeft, focusRight)
+    /// - Returns: The next window in the specified direction, or `nil` if no suitable window is found
+    static func nextWindow(from currentWindow: Window?, in direction: WindowDirection) -> Window? {
```
mrkai77: I feel like "next" is contradictory with the term "direction" as they mean two different things in screen switching. Also, since the direction cannot be *any* direction, it would make sense to check if `direction.focusEdge` exists *before* calling this function:

```suggestion
    static func directionalWindow(from currentWindow: Window?, edge: Edge) -> Window? {
```

(note that this makes it more similar to the parameters of `directionalScreen` in ScreenUtility)

MrKai77 (changes requested):

Loop/Window Management/Window/WindowUtility+FocusNavigation.swift
cipher-shad0w: I set the logger in WindowUtility to static let and then I used the same logger in FocusNavigation

Loop/Window Management/Window/WindowUtility+FocusNavigation.swift
```diff
@@ -0,0 +1,112 @@
...
+    /// Finds the next window to focus in the specified direction.
+    /// - Parameters:
+    ///   - currentWindow: The currently focused window to navigate from, or nil to navigate from screen center
+    ///   - edge: The direction to search for the next window (leading, trailing, top, bottom)
+    /// - Returns: The next window in the specified direction, or `nil` if no suitable window is found
+    static func directionalWindow(from currentWindow: Window?, edge: Edge) -> Window? {
+        let allWindows = windowList()
+
+        // If no current window, navigate from screen center
+        if currentWindow == nil {
```
MrKai77: Instead, use `if let currentWindow` and move the code that currently comes after this statement inside that block, so the force unwrap on line 77 isn’t needed. Then, place this logic inside an `else` block.

Loop/Window Management/Window/WindowUtility+FocusNavigation.swift
```diff
@@ -0,0 +1,112 @@
...
+                .filter { window in
+                    isInDirection(window.frame, from: screenCenter, edge: edge)
+                }
+                .min { window1, window2 in
+                    screenCenter.distance(to: window1.frame.center) < screenCenter.distance(to: window2.frame.center)
+                }
+
+            if let nextWindow {
+                let nextWindowName = nextWindow.nsRunningApplication?.localizedName ?? nextWindow.title ?? "<unknown>"
+                logger.info("[FocusNavigation] Found window to focus in direction \(String(describing: edge)): \(nextWindowName)")
```
MrKai77: I know the rest of the codebase hasn't been fully updated to use this, but prefer:
```suggestion
                logger.info("[FocusNavigation] Found window to focus in direction \(String(describing: edge)): \(nextWindow.debugDescription)")
```

Loop/Window Management/Window/WindowUtility+FocusNavigation.swift
```diff
@@ -0,0 +1,112 @@
...
+            return nil
+        }
+
+        // Use the generic directional navigation from DirectionalNavigationUtility
+        if let nextWindow = DirectionalNavigationUtility.directionalItem(
+            from: currentWindow!,
+            in: otherWindows,
+            edge: edge,
+            canRestartCycle: true,
+            frameProvider: { $0.frame }
```
MrKai77:
```suggestion
            frameProvider: \.frame
```

Loop/Utilities/DirectionalNavigationUtility.swift
```diff
@@ -0,0 +1,150 @@
...
+            }
+            .min {
+                distanceBetweenFrames(for: edge, current: currentFrame, other: frameProvider($0)) <
+                    distanceBetweenFrames(for: edge, current: currentFrame, other: frameProvider($1))
+            }
+    }
+
+    // Find overlapping items in the same axis
+    private static func overlappingItems<T>(
+        from _: T,
```
MrKai77: This can be removed:

```suggestion

```

MrKai77 (approved):

PR #873, opened 2025-11-02
🐞 Fix recording of arrow key keybindings

### LINKED ISSUE #871: 🐞 Can't assign keybinding trigger + arrow
### Bug Description

When I try to assign a keybinding to trigger + arrow, Loop instead assigns trigger + globe arrow

### Steps To Reproduce

1. Settings → Keybinds
2. Click binding to reassign
3. Type ↓ 

### Expected Behavior

Binding shows: [trigger] + ↓ 

### Actual Behavior

Binding shows: [trigger] + 🌐 ↓ 

### Screenshots

Image

### macOS Version

Sequoia 15.7.1 (24G231)

### Loop Version

1.4.0 (1562)

### Additional Context

My keyboard doesn't have 🌐 key, so the binding does nothing for me.

### Final Checks

- [x] I can help with fixing or developing this issue (tick if you can help).

humblehacker:
Fixes the inability to record an arrow keybinding without also including the globe/fn key by removing the function key from the captured modifier keys if the key is a "special" key.

### DIFF
--- a/Loop/Settings Window/Settings/Keybindings/Keybind Recorder/Keycorder.swift
+++ b/Loop/Settings Window/Settings/Keybindings/Keybind Recorder/Keycorder.swift
@@ -133,8 +133,12 @@ struct Keycorder: View {
         let currentKeys = selectionKeybind + [event.keyCode]
             .map { $0.baseKey(flags: event.modifierFlags) }
 
-        let flags = CGEventFlags(cocoaFlags: event.modifierFlags)
+        var flags = CGEventFlags(cocoaFlags: event.modifierFlags)
 
+        if event.keyCode.isFnSpecialKey {
+            flags.remove(.maskSecondaryFn)
+        }
+        
         // Filter out trigger keys from flags
         let validModifiers = flags.keyCodes.filter {
             !Defaults[.triggerKey]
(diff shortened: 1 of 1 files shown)


MrKai77:
Forwarding your previous message here for context:
> Just pressing the arrow key. I did some debugging and it seems that the function modifier always accompanies the arrow key in the key event - at least on the two macs I've tried it on. After some digging, I ended up with a fix that seems to work, though I admit I don't fully understand it. Here's a draft PR with the small change: https://github.com/MrKai77/Loop/pull/873. Maybe the change makes more sense to you?

---

Yes, macOS can be really weird about the Function (i.e. Globe) key showing up as “pressed” alongside unrelated keys, so your fix does actually make sense. I’ve spent countless hours trying to find a more logical workaround myself, but nothing fit well with Loop once its other requirements were taken into account. (the other option was to listen to hardware-level key events)

I’m not entirely sure why Apple marks certain keys with the Function modifier flag; it definitely feels inconsistent. There’s also no documentation explaining this behavior, which doesn’t help.

Also, to add to this inconsistency, I wasn’t able to reproduce the original issue tied to this PR; likely because I’m using an Apple keyboard. Third-party keyboards could very well be treated differently by macOS.

You can see the full list of (known) affected keys inside `CGKeyCode.isFnSpecialKey`'s body (which you used). My guess is that this behavior exists because the Function key was never originally designed to be a true, general-purpose modifier, at least not until recent years, when Apple started using it for system-level window management shortcuts and a few other (quite) obscure bindings. Even then, its use as a modifier still seems to be largely limited to Apple’s own ecosystem.

---

So given all that, it makes sense to merge this PR :)
If you’d like to rebase and mark it as ready, I’ll be happy to get it merged!

humblehacker:
Thanks for the explanation! Makes sense, inasmuch as undocumented idiosyncratic library code can make sense :)

PR #875, opened 2025-11-02
🐞 Fix window animation lag during screen transitions

cipher-shad0w:
## Description

Fixes a race condition in `WindowTransformAnimation` that caused visible stuttering and incorrect window positioning during animations, particularly when moving windows between screens with different resolutions.

## Problem

The animation loop was updating `lastWindowFrame` using `window.frame`, which internally calls the Accessibility API to get the current window position and size. However, the Accessibility API is **asynchronous** - when we set `window.position` or `window.size`, the actual window update happens later in the target application.

This caused a timing issue:
1. Frame N: Set `window.position = newFrame.origin` (async request sent)
2. Frame N: Read `lastWindowFrame = window.frame` (returns **stale** value from Frame N-1)
3. Frame N+1: Bounds checking uses incorrect `lastWindowFrame` values
4. Result: Window jumps/stutters or is positioned incorrectly

## Solution

Changed line 89 in `WindowTransformAnimation.swift`:

```swift
// Before (buggy):
lastWindowFrame = window.frame  // Async API call, returns stale data

// After (fixed):
lastWindowFrame = newFrame      // Synchronous local variable, always current
```

Using the locally calculated `newFrame` ensures the bounds checking logic always uses the intended frame values, eliminating the race condition.

### DIFF
--- a/Loop/Window Management/Window Manipulation/WindowTransformAnimation.swift
+++ b/Loop/Window Management/Window Manipulation/WindowTransformAnimation.swift
@@ -86,7 +86,7 @@ final class WindowTransformAnimation: NSAnimation {
                 window.size = newFrame.size
             }
 
-            lastWindowFrame = window.frame
+            lastWindowFrame = newFrame
 
             if currentProgress >= 1.0 {
                 WindowTransformAnimation.currentAnimations[window.cgWindowID] = nil


cipher-shad0w (commented):
I found the bug while fixing issue #713

MrKai77:
This is not a bug, and it is intended behavior. Some windows have size constraints, preventing them from being able to resize to the requested size. By using the window's latest frame rather than the local variable, we can take this into consideration. Here's an example of how it affects the animation:

With your changes:
<div><video controls src="https://github.com/user-attachments/assets/cb73f7ca-cb45-4e26-ad0e-7ea372ae59e7" muted="false"></video></div>
As you can see, it stutters at the very end of the animation when resized to the right half, since the window's minimum size isn't taken into consideration.

---

Without your changes:
<div><video controls src="https://github.com/user-attachments/assets/7651e009-7bf7-4cb5-99f8-49a720329ba4" muted="false"></video></div>

cipher-shad0w:
Ah I understand the issue, thx for the explanation

PR #893, opened 2025-11-18
✨ Lightweight `IconView`

MrKai77:
For some time, the SwiftUI-based `IconView` has occasionally caused issues. The most notable problem was that it could lag one state behind after a user updated a selection in the keybinds tab. Other issues included stuttering animations and high rendering costs, although the latter could potentially have been optimized within SwiftUI itself.

By switching to a dedicated `NSView` backed by `CALayers` for rendering icons, we gain more control over when the view updates and benefit from a more lightweight rendering backend. This allows Loop to focus its resources on higher-priority tasks in the settings window.

### DIFF
--- a/Loop/Settings Window/Settings/Keybindings/Item View/DirectionPickerView.swift
+++ b/Loop/Settings Window/Settings/Keybindings/Item View/DirectionPickerView.swift
@@ -8,7 +8,7 @@
 import SwiftUI
 
 struct DirectionPickerView: View {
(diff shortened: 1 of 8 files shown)


SenpaiHunters (approved):
I've reviewed each file and noticed that sizing changes were made, for example, from 12 to 14. I assume this is in accordance with macOS 26 or for another reason. Visually, I don't see a significant change. Regardless, everything is accepted, and I hope this will resolve some issues we've been experiencing!

PR #903, opened 2025-11-25
✨ Introduce SkyLightToolBelt, ability to use window corner radius in preview

### LINKED ISSUE #898: ✨ Adaptive Corner Radius Rendering Based on Window Style

I have been considering this idea, and I think it would be aesthetically pleasing if Loop could render preview animations based on the type of window being moved. In macOS 26, different window styles have distinct corner radii, so applying a fixed corner radius does not work consistently.

While I understand that even macOS itself does not handle this by default, it might be possible to implement a condition based on the bundle ID or the window ID.  

If Loop has access to window properties, whether through the Accessibility framework or another mechanism, it could potentially render frames in a variable way depending on the window style of each application.

This would not only provide a more visually coherent result, but also further elevate Loop beyond the native window management capabilities of macOS.

### Screenshots

_No response_

### Additional Context

_No response_

### Final Checks

MrKai77:
- Consolidates all SkyLight interactions into a new `SkyLightToolBelt` file. Several SkyLight functions were previously scattered across the codebase, so they’ve been centralized with wrapper functions for cleaner and easier usage.
- Adds a new feature for the preview window to use the selected window’s actual corner radius instead of a user-defined value.


SenpaiHunters (approved):

MrKai77:
Note: Once a few issues are fixed, I’ll reopen a new PR based on changes from this branch.

PR #905, opened 2025-11-27
🐞 Reset `ListenEvent` when requesting Accessibility access

MrKai77:
- Runs `tccutil reset ListenEvent com.MrKai77.Loop` when resetting permissions
- Makes sure that `WindowDragManager` doesn't start its event monitors if permissions haven't been granted


SenpaiHunters (approved):
This was such a hassle to find and fix, but this PR now resolves the issues!

PR #918, opened 2025-12-08
💄 Luminare UI updates

MrKai77:
- Replaces Nucleo icons with SF Symbols for a more native macOS appearance
- Adds colored sidebar icons to more closely match system apps
- Enhances light mode such that section backgrounds are now brighter than their surroundings (inspired by Alcove)
- Buttons now feature a distinct border to improve visibility and clickability
- Keybind search now supports fuzzy matching
- Removes many UI animations for a snappier experience
- Updates `string.replacingOccurrences(of:)` to `string.replacing()`
- Improves UI alignment for non-English localizations


Loop/Settings Window/SettingsTab.swift
```diff
@@ -39,15 +76,15 @@ enum SettingsTab: LuminareTabItem, CaseIterable {
 
     var image: Image {
         switch self {
-        case .icon: Image(.squareSparkle)
-        case .accentColor: Image(.paintbrush)
+        case .icon: Image(systemName: "sparkles.2")
```
SenpaiHunters: `No symbol named 'sparkles.2' found in system symbol set.` This is only for macOS 26 by the looks of it; however, 'sparkle' itself is available.

SenpaiHunters (changes requested):
That is the only item I can see that will cause issues; the rest looks good. I tested it in light mode (eww) and dark mode. All toggles activate, all switches work, and adding/removing doesn't crash. Everything seems good apart from the small non-included sparkle.

Loop/Settings Window/SettingsTab.swift
mrkai77: Will use `sparkles` which seems to be available pre-Tahoe

PR #943, opened 2026-01-05
✨ Precompute padding

### LINKED ISSUE #706: 🐞 `Maximise Height` + Simple, 10px padding causes window of maximised height-window to shrink in width (in relation to original window).
### Bug Description

I have been having an issue for a while where maximise height causes a different width for the window. I believe the cause is that it adds an extra padding to the width of the orginal window, while the width should remain the same. Practically I think what happens is that the original window has +10px padding, and when reframing to maximise height it adds another +10px. 

I posted a video [here](https://github.com/MrKai77/Loop/pull/605#issuecomment-2476495649).

_My settings:_
Radial menu → `off`
Preview → `off`
Include padding → `Simple`, `10px`
Respect stage manager → `off`
Disable cursor interaction → `off`

### Steps To Reproduce

1. Activate a difined keybinding with a certain width.
2. Activate a keybinding that should maximise the height of that original window.
3. See that the window has shrunk in width, while maximised in height.

### Expected Behavior

The expected behaviour is that when `Maximise Height` of a window, it does not change other characteristics of that window, mainly the width.

### Actual Behavior

When `Maximise Height` it adds another 10px padding to the original window (this is an assumption).

### Screenshots

_No response_

### macOS Version

Version 15.4 Beta (24E5222f)

### Loop Version

🧪 1.3.0 (1451)

### Additional Context

_No response_

### Final Checks

MrKai77:


SenpaiHunters (changes requested):
All code changes work, only request is running `swiftformat .` before committing!

PR #952, opened 2026-01-05
✨ Check for minimum macOS version in future updates

MrKai77:


SenpaiHunters (approved):
This works; testing to add a small debug, both return 13.0. The required and current macOS version is X.

PR #955, opened 2026-01-06
✨ Allow calculated mouse position to exceed screen bounds

MrKai77:
Allows the calculated mouse position to exceed screen bounds, so that users can access directional radial menu actions despite the mouse being pinned at the screen edge.


SenpaiHunters (approved):
It workie

Issue #260, opened 2024-03-20
✨ Generic keyboard shortcut to trigger Loop


delgross:
### Feature Request Description

I am trying to set up a system to trigger the program with a trackpad gesture.  It would very helpful if I could trigger the menu with a keyboard shortcut and not a modifer key

### Final Checks

- [X] My issue title is descriptive
- [X] This is a single feature request (multiple feature requests should be reported individually)

liangch:
I am pleased with keyboard triggering from skhd when using yabai.  Hope this feature is available here.

ortonomy:
I know we don't like +1s, but +1. This is quite old but desperately needed. I'll come back in a year to see if this gets fixed

PR #963, opened 2026-01-07
✨ Allow select keybinds to ignore the trigger key

cipher-shad0w:
#260


cipher-shad0w:
There’s just one problem right now — currently, you can set any single key as a trigger for an action, for example just a single letter. What should I check for? A minimum of 2 keys? And certain keys that must be included so that the custom trigger key can be used?

Loop/Window Management/Window Action/WindowAction.swift
```diff
@@ -94,6 +96,7 @@ struct WindowAction: Codable, Identifiable, Hashable, Equatable, Defaults.Serial
     // Generic Properties
     var direction: WindowDirection
     var keybind: Set<CGKeyCode>
+    var bypassTriggerKey: Bool = false
```
mrkai77: Make this property optional; otherwise, all users will lose their existing keybinds since their keybinds don't include it.

MrKai77 (commented):
> There’s just one problem right now — currently, you can set any single key as a trigger for an action, for example just a single letter. What should I check for? A minimum of 2 keys? And certain keys that must be included so that the custom trigger key can be used?

I'd say a minimum of at least one modifier key and one non-modifier key, and then a maximum of around 5 keys in total (e.g. 4 modifiers, one key).

Loop/Window Management/Window Action/WindowAction.swift
cipher-shad0w: ah thats why i lost all my keybinds

cipher-shad0w:
> I'd say a minimum of at least one modifier key and one non-modifier key, and then a maximum of around 5 keys in total (e.g. 4 modifiers, one key).

Does Loop has anything implemented to check if sth is a modifier key or not?

MrKai77:
There is `CGKeyCode.isModifier`. Additionally, it likely makes sense to map these via `CGKeyCode.baseModifier` to eliminate the left/right key code distinction.

MrKai77:
@cipher-shad0w do you have any updates on this?

cipher-shad0w:
Oh yeah sry, I've implemented everything you ask for and fixed everything you commented

MrKai77 (approved):
I made a few changes that fixed some bugs on my end and also optimized parts of both your code and the existing code. Everything works perfectly now, thank you!

PR #966, opened 2026-01-08
💄 Change "Restart to complete" to "Relaunch to complete" in update window

kylefmohr:
Hi! Thank you for this wonderful app.

I don't know if anybody else made the same mistake as me, but the first few times I saw Loop say, after an update, "Restart to complete", I thought it was going to restart my whole computer. 

I feel that this wording is a bit clearer on the actual behavior of the app update process.


Loop/Updater/UpdateView.swift
```diff
@@ -86,7 +86,7 @@ struct UpdateView: View {
                         }
 
                         let tenSpaces = "          " // This helps with alignment for the animation once the update finishes
-                        Text(isInstalling ? tenSpaces : readyToRestart ? NSLocalizedString("Restart to complete", comment: "") : NSLocalizedString("Install", comment: ""))
+                        Text(isInstalling ? tenSpaces : readyToRestart ? NSLocalizedString("Relaunch Loop to complete", comment: "") : NSLocalizedString("Install", comment: ""))
```
mrkai77:
```suggestion
                        Text(isInstalling ? tenSpaces : readyToRestart ? NSLocalizedString("Relaunch to complete", comment: "") : NSLocalizedString("Install", comment: ""))
```
Should be enough.

MrKai77 (changes requested):
Looks good overall. Just a couple of notes:
- Please build the app in Xcode at least once to refresh the localizations file.
- It would also be great to migrate the old localization key to the new one so existing translations aren’t lost.

Loop/Updater/UpdateView.swift
kylefmohr: Thank you! I'm inexperienced with using localizations in Xcode. I will take care of both of these momentarily.

kylefmohr:
> Looks good overall. Just a couple of notes:
> 
> * Please build the app in Xcode at least once to refresh the localizations file.
> * It would also be great to migrate the old localization key to the new one so existing translations aren’t lost.

Thank you! I'm inexperienced with using localizations in Xcode. I will take care of both of these momentarily.

MrKai77 (approved):

PR #1011, opened 2026-01-21
🐞 Reduce drag event overhead

### LINKED ISSUE #1001: 🐞 Loop ≥1.4 lags and sometimes holds inputs hostage in Minecraft
### Bug Description

Just updated Loop to 1.4.1 today and noticed some weird behavior while playing Minecraft. When moving the mouse while holding down LMB and W/A/S/D, the mouse will lag around significantly as if something is holding the input data hostage until a bit later, making the camera snap around instead of moving smoothly. Sometimes the W/A/S/D key and/or LMB will also be "stuck" down despite already being released, making the character move continuously in a direction and/or breaking blocks continuously without any user input. This does not happen in v1.3.

### Affected Scope

Performance

### Steps to Reproduce

1. Load Minecraft up and set a very high FPS limit (in my case w/ shaders on, 250FPS or Unlimited both work, the higher the difference between the actual fps and the fps limit the higher chance it happens, sometimes switching between 250 and ∞ triggers the bug immediately)
2. Hold down LMB and move the mouse around
3. Keep LMB held down and start moving randomly with WASD
4. Notice how the camera starts spazzing out and lag-snaps to different positions trying to catch up with the mouse movement

### Reproducibility

Often (≥70%)

Expected: the cursor and the camera to move around smoothly with no hiccups

Actual: the camera lags around and sometimes inputs are held hostage

https://github.com/user-attachments/assets/eff85a1d-fed4-43e0-be55-b8e43e69fe58

https://github.com/user-attachments/assets/cff7c9e6-ce5c-47f9-b894-d8f20aca1300

also notice the second clip at ~0:06 i did not press s but the s seemingly got delayed from minecraft to raycast turning the search from loop to loops, and later in the same clip raycast sometimes didnt open immediately while the bug is happening

### Severity

Major (workaround exists)

### macOS Version

Tahoe 26.2

### Loop Version

Version 1.4.1 (1677)

No

### Additional Context

Happens on both Logitech M720 (bluetooth and unifying) and trackpad so not a hardware issue  
Can not reproduce when using Loop 1.3 or when Loop is not running  
The issue is also present on the [oldest gh actions artifact that is still downloadable](https://github.com/MrKai77/Loop/actions/runs/18620534178)  
Switching between 250fps and unlimited usually activates the bug right after

### Final Checks

- [ ] I can help with further investigation.

manuthebyte:
#1001 

### tldr
- Prevents global drag monitoring from doing work unless snapping/restore/stash features are active.
- Stops repeated window-resolution attempts during non-window drags (e.g., games).
- Resets drag state on mouse up to avoid stale drag tracking.

### Tests
- I have run my tests on my NMS instance, works as expected now and doesn't bug anymore.

### DIFF
--- a/Loop/Core/WindowDragManager.swift
+++ b/Loop/Core/WindowDragManager.swift
@@ -20,6 +20,8 @@ final class WindowDragManager {
     private var draggingWindow: Window?
     private var initialWindowFrame: CGRect?
     private var direction: WindowDirection = .noAction
+    // Avoid repeated window resolution attempts during a non-window drag (e.g. in games).
+    private var didFailToResolveDraggedWindow: Bool = false
 
     private let previewController = PreviewController()
 
@@ -31,6 +33,13 @@ final class WindowDragManager {
 
     private var currentMousePosition: CGPoint {
         NSEvent.mouseLocation.flipY(screen: NSScreen.screens[0])
+    }
+
+    // Avoid running global drag logic unless a feature actually depends on it.
+    private var shouldMonitorDragActions: Bool {
+        Defaults[.windowSnapping] ||
+            Defaults[.restoreWindowFrameOnDrag] ||
+            !Defaults[.stashManagerStashedWindows].isEmpty
     }
 
     @MainActor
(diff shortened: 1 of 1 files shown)


MrKai77 (approved):
This looks good, and it makes sense that this would cause lag. Thank you!

Loop/Core/WindowDragManager.swift
```diff
@@ -135,11 +145,10 @@ final class WindowDragManager {
                 }
             }
 
-            self.previewController.close()
-            self.draggingWindow = nil
-
```
mrkai77, line 140: Thanks for also removing these duplicate calls :)

manuthebyte: no problem! :)

PR #1012, opened 2026-01-21
🐞 Fix rapid keypress mismatch

zenangst:
Use `CGEvent.keyCode` as a fallback if the `actionKeys` fail to match any `WindowAction`.

Fixes an issue where the user invokes keyboard shortcuts in rapid succession, and if they are fast enough, the list of action keys can include multiple values.


MrKai77 (approved):
Works great, thank you!

PR #1038, opened 2026-02-02
✨ Ability to launch Loop hidden, add bottom stash option

### LINKED ISSUE #962: ✨ Support stash window to top/bottom of display
### Feature Request Summary

Currently, Loop only support stash window to left/right of display, could add support to also stash window to top/bottom of display?

Image

### Problem / Opportunity & Desired Solution

Because in my current display setup, if I stash to the left, I will trigger the macOS stage manager "recent apps":

Image

If I stash right, my mouse would jump to my iPad because of the universal control feature:

Image

So I would try the stash window to the top or bottom of the display, these directions may be safe areas.

### Alternatives Considered

_No response_

### Visuals or References

_No response_

### Additional Context

_No response_

### Priority

Low (nice to have)

### Final Checks

- [x] This is a single feature request (multiple feature requests should be reported individually).
- [x] I have looked to see if this is a duplicate of another feature request.
- [ ] I can help with research or testing.
- [ ] I can help code this feature.

### LINKED ISSUE #1031: ✨ Startup Loop in the background
### Feature Request Summary

To work around an illustrator Bug I'm quitting and starting Loop using apple shortcuts, but every time it starts it now shows its window. Please allow starting of loop without showing the window (Or fix the Illustrator ignoring bug, either is fine for me really - PS, LOVE THE "Check for updates" microcopy! <3 ;)

### Problem / Opportunity & Desired Solution

Allow starting hidden

### Alternatives Considered

Apple shortcuts, quite hacky!

### Visuals or References

_No response_

### Additional Context

_No response_

### Priority

Medium (important but not blocking)

### Final Checks

- [x] This is a single feature request (multiple feature requests should be reported individually).
- [x] I have looked to see if this is a duplicate of another feature request.
- [ ] I can help with research or testing.
- [ ] I can help code this feature.

SenpaiHunters:
This update aims to add two minor features requested by users: the ability to start up Loop without opening a window and to add stash windows to the top and bottom of the display.

A note for both: first, with Loop being hidden, if both the menu bar and the app on launch are hidden, and you're in development mode, i.e., opening it in Xcode, you will NOT be able to launch the window. You'll need to disable this before starting, or alternatively turn on the icon. If none of these are available, you can simply

```sh
defaults write com.MrKai77.Loop startHidden -bool false
```

Regarding window stashing, for stashing, the calculation is done, i.e., the math. If you have two window sizes, both previewing at 80%, it will remove the windows at the top and bottom. My recommendation here is to set the window width to be half of the maximum. For example, if you have two windows, instead of both being 50%, set them to 40-45% (given that both are stashed on the right-hand top and bottom of the screen).

### DIFF
--- a/Loop/App/AppDelegate.swift
+++ b/Loop/App/AppDelegate.swift
@@ -31,7 +31,8 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
             await Defaults.iCloud.waitForSyncCompletion()
         }
 
-        if !launchedAsLoginItem {
+        // Show settings window only if not launched as login item AND startHidden is disabled
+        if !launchedAsLoginItem, !Defaults[.startHidden] {
             SettingsWindowManager.shared.show()
         } else {
             // Closing also hides the dock icon if needed.
--- a/Loop/Extensions/Defaults+Extensions.swift
+++ b/Loop/Extensions/Defaults+Extensions.swift
@@ -41,6 +41,7 @@ extension Defaults.Keys {
 
     // Behavior
     static let launchAtLogin = Key<Bool>("launchAtLogin", default: false, iCloud: true)
+    static let startHidden = Key<Bool>("startHidden", default: false, iCloud: true)
     static let hideMenuBarIcon = Key<Bool>("hideMenuBarIcon", default: false, iCloud: false)
     static let animationConfiguration = Key<AnimationConfiguration>("animationConfiguration", default: .snappy, iCloud: true)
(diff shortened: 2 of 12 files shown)


SenpaiHunters:
007567-02@2x

I am not sure what the linter is failing for; I can't see it on my side...

MrKai77 (approved):
Looks good! I removed the top stash action since the menu bar often interfered with it (added a comment on that commit), causing odd behavior. I also tweaked the stash action configuration window so the top/center actions are now greyed out.

PR #1048, opened 2026-02-16
✨ Add 'Ignore notch' option for consistent top padding across screens

### LINKED ISSUE #691: ✨ Different padding for notched internal monitor and external monitor

I use SketchyBar with the default menu bar hidden in MacOS.

When I'm on my laptop the external bar is calculated from the notch, but SketchyBar is behind the notch. 
I would like to set different padding / external bar for my screens with a notch and those without a notch.

### Screenshots

_No response_

### Additional Context

_No response_

### Final Checks

kjvdven:
Adds an defaults write option "Ignore notch". When enabled, the notch height is subtracted from the top padding on notched displays, so the effective distance from the screen's top edge matches non-notch displays.

Users can enable with:
```
defaults write com.MrKai77.Loop ignoreNotch -bool true
```

Reset with: 
```
defaults delete com.MrKai77.Loop ignoreNotch
```

Users with an external menu bar (e.g. SketchyBar) and the system menu bar hidden experience inconsistent top padding between notched and non-notch displays. On a notched MacBook, `NSScreen.visibleFrame` reserves space for the notch (~32pt), and the external bar padding stacks on top of that — resulting in more top gap than on an external monitor.

When "Ignore notch" is enabled, `applyToBounds` subtracts `screen.menubarHeight` from `totalTopPadding` (clamped to 0), compensating for the notch reserve. Backward-compatible: custom `Codable` init defaults `ignoreNotch` to `false` for existing configs.

Fixes #691

## How has this been tested?

Tested on a MacBook Pro (notch) with an external monitor, system menu bar hidden, SketchyBar as external bar with 40px external bar padding configured in Loop.

- [x] External monitor: top padding remains 40px as configured
- [x] `defaults write com.MrKai77.Loop ignoreNotch -bool true`: top padding matches external monitor (40px from screen top)
- [x] `defaults delete com.MrKai77.Loop ignoreNotch`: top padding is 40px + notch height (existing behavior)
- [x] Existing configs without the new field load correctly (defaults to false)

macOS 15, Loop built from develop branch.

## Checklist:

- [x] I have made corresponding changes to the documentation if applicable

This change was developed with assistance from Claude (Anthropic). The approach (subtracting `screen.menubarHeight` from top padding when the toggle is enabled) was designed collaboratively — the math and edge cases were worked through together, and all code was reviewed, understood, and tested by the contributor on a real notch + external monitor setup.

### DIFF
--- a/Loop/Core/LoopManager.swift
+++ b/Loop/Core/LoopManager.swift
@@ -319,7 +319,7 @@ extension LoopManager {
 
                         let adjustedBounds = PaddingConfiguration
                             .getConfiguredPadding(for: currentScreen)
-                            .applyToBounds(currentScreen.cgSafeScreenFrame)
+                            .applyToBounds(currentScreen.cgSafeScreenFrame, screen: currentScreen)
 
                         let proportionalSize = CGRect(
                             x: (currentFrame.minX - adjustedBounds.minX) / adjustedBounds.width,
--- a/Loop/Extensions/Defaults+Extensions.swift
+++ b/Loop/Extensions/Defaults+Extensions.swift
@@ -106,6 +106,12 @@ extension Defaults.Keys {
     /// Reset with `defaults delete com.MrKai77.Loop paddingMinimumScreenSize`
     static let paddingMinimumScreenSize = Key<CGFloat>("paddingMinimumScreenSize", default: 0, iCloud: true)
 
+    /// Ignore the notch height when calculating top padding, so the effective
+    /// distance from the screen top matches non-notch displays.
+    /// Adjust with `defaults write com.MrKai77.Loop ignoreNotch -bool true`
+    /// Reset with `defaults delete com.MrKai77.Loop ignoreNotch`
+    static let ignoreNotch = Key<Bool>("ignoreNotch", default: false, iCloud: true)
+
     /// Snap threshold for window snapping, defined in points.
     /// Adjust with `defaults write com.MrKai77.Loop snapThreshold -float x`
     /// Reset with `defaults delete com.MrKai77.Loop snapThreshold`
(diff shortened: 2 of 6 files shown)


MrKai77 (changes requested):
Thanks for the PR! Overall, this works well and the logic is correct. However, I think `ignoreNotch` belongs as a global setting rather than on PaddingConfiguration.

A couple of reasons:
1. It's a hardware/setup concern, not a padding preference. Whether to compensate for the notch is a property of the user's display setup, not a specific padding configuration. Conceptually it fits better as a global toggle (e.g. `Defaults[.ignoreNotch]`). Given that this is a fairly niche setup (hidden menu bar + external bar + notch display), I'd lean toward making it a `defaults write` option rather than a visible UI toggle. This also avoids adding complexity to the settings UI for something most users won't need.
2. It avoids the custom Codable conformance. Right now the PR replaces the auto-synthesized Codable with a manual `init(from:)` just to default `ignoreNotch` to false for existing configs. Pulling the flag out into a separate Defaults key lets `PaddingConfiguration` go back to synthesized conformance, keeping it clean.

kjvdven:
Thnx for the review, 

1. I changed the code to work with a `default write` instead of a UI toggle.
1. Removed the custom Codable conformance

MrKai77 (approved):
Thank you!

cyb3rko:
@kjvdven Does that still work for you?  
I can not get it to work somehow :(

kjvdven:
> @kjvdven Does that still work for you? I can not get it to work somehow :(

Have you set the variable in the terminal? 

```bash
defaults write com.MrKai77.Loop ignoreNotch -bool true
```

cyb3rko:
> Have you set the variable in the terminal?

I have, but I can not see any changed behaviour.  
Any other setting that is required for that option to work?

MrKai77:
@cyb3rko could you check if the macOS window manager is enabled in the Advanced tab? That setting can interfere with this feature. Also, make sure you’re running the latest development version :)

cyb3rko:
Well, call me stupid.  
I just realized that it's not included in 1.4.2, works now on pre-release version 👍.

Thanks to both of you

PR #1063, opened 2026-03-05
✨ Use privileged updater helper if needed

MrKai77:
This branch started as a fix for updater failures in non-admin sessions, but it ended up being a full updater hardening pass!

The main change is a new privileged helper, `LoopUpdaterHelper`, that Loop can use during update installation when the current user session cannot write to the install location. Previously, this could cause the update to fail outright.

Security was a huge focus here. The helper is intentionally strict:
- Loop and helper communicate over XPC with a shared protocol.
- The helper validates the caller using bundle identity + code-signing requirement checks.
- The helper does not accept arbitrary file paths from Loop.
- Instead, Loop sends a rollback token, and the helper derives/validates all paths itself from trusted context.
- Rollback IDs are validated (format/length/traversal protection), and derived paths must remain inside trusted staging/rollback roots.
- The helper validates bundle signatures before privileged swap/restore operations.
- The helper only allows one active privileged connection at a time.
- Ownership is explicitly managed: rollback copy gets client uid/gid ownership, installed app gets root ownership.

Installation flow changes in Loop:
- Added `UpdaterAuthorizationCoordinator` to manage privileged readiness, Authorization Services rights, one-shot `SMJobSubmit`, helper calls, timeout handling, and `SMJobRemove` cleanup.
- Added permission-state routing (`writable`, `needsElevation`, `notWritableNoElevationPossible`) so install behavior is explicit.
- Privileged flow now stages and verifies after authorization is granted, then performs privileged atomic swap via helper.
- Added privileged failure reconciliation + restore path and ownership verification for the installed app.
- If privileged install cannot proceed, Loop now offers a user-facing fallback to install in `~/Applications` instead of hard-failing.

Backup/rollback/indexing behavior changes:
- Replaced old path helper usage with shared `LoopSupportPaths`.
- Added shared staging/backup/rollback path model under `~/Library/Application Support/Loop`.
- Rollback working copies now use `Rollback.noindex` to avoid Spotlight/app-launcher indexing.
- Rollback snapshots are archived into `.zip` backups after swap.
- Backup manager now:
  - stores backups as `.zip`,
  - purges legacy non-zip backup artifacts,
  - performs best-effort old-backup cleanup when size exceeds threshold,
  - can prepare restore by unzipping the latest archive into staging.

Performance and reliability improvements:
- Checksum verification is now streaming SHA-256 (1 MiB chunks) instead of loading full files into memory.
- Download hash checking now includes explicit file existence/readability guards and byte-count logging.
- Installer code-sign validation moved from shelling out to `codesign` to Security framework APIs.
- Directory size calculation was centralized in a `FileManager` extension and reused where needed.


Loop/Updater/UpdaterAuthorizationCoordinator.swift
```diff
@@ -0,0 +1,341 @@
+import Foundation
```
SenpaiHunters, line 8: Add code top comment, ie
```
//
//  UpdaterAuthorizationCoordinator.swift
//  Loop
//
//  Created by Kai Azim on 2026-03-05.
//
```

LoopUpdaterHelper/PrivilegedInstaller.swift
```diff
@@ -0,0 +1,498 @@
...
+            try fileManager.moveItem(at: stagedURL, to: currentURL)
+            log.success("Moved staged app into current location")
+            try applyRootOwnershipRecursively(at: currentURL)
+            log.success("Applied root ownership to installed app")
+        } catch {
+            log.warn("Swap failed after backup move; attempting rollback")
+            try? fileManager.removeItem(at: currentURL)
+            try? fileManager.moveItem(at: backupBundleURL, to: currentURL)
+            try? applyRootOwnershipRecursively(at: currentURL)
+            log.success("Rollback to backup completed")
```
SenpaiHunters: For this, when we're logging the `log.success("Rollback to backup completed")` unconditionally, as rollback actions are using `try?`, it might silently fail. If the rollback fails, surface that clearly in the logs or return an error.

SenpaiHunters (changes requested):
Remember to also include the README updates 😄.

If we're fine overall with the `try?` log, then it can be accepted as is.

PR #1064, opened 2026-03-05
✨ Trackpad gestures

### LINKED ISSUE #138: ✨ Trackpad gestures to trigger Loop

mrkai77:
Adds trackpad gestures!

With this PR, Loop will now include a new package dependency called _Subsurface_. This is a Swift package that provides a clean interface to the private `MultitouchSupport` framework. Here’s a small demo of trackpad gestures in action (although this is actually quite outdated, the current functionality is much more refined!). The plan is for it to trigger only when the cursor is over the title bar, similar to Swish:
<div><video controls src="https://github.com/user-attachments/assets/343ab029-bd3e-4234-828e-26361115e3a5" muted="false"></video></div>


SenpaiHunters:
Remember to update the readme table, to now show that we support trackpad gestures! 🎉

JaceThings:
Goated

AlejandroVolkova:
Wow, this is awesome! It’s exciting to see it coming together! I just had a couple of thoughts after watching the demo. In the video, I noticed that windows move and resize instantly with the gestures. Would it be possible to instead show the preview animation before the actual move? Especially for circle gestures, previewing the desired position before the actual window move would be helpful and could make the overall experience feel smoother (essentially the same behavior you get when using the trigger key or window snapping.)

I was also wondering if custom gestures might be supported. For example, being able to assign a gesture to maximize or minimize windows. Personally, I’d love to disable Mission Control and Exposé so I could use a four-finger drag up/down gesture for maximize/minimize instead.

And finally, will haptic feedback be part of this feature? It could add a really nice touch to the interaction.I know this is still under development, so please take these just as friendly suggestions in case they’re helpful 😅 Thanks for making this possible!

MrKai77:
Thanks @AlejandroVolkova!

> In the video, I noticed that windows move and resize instantly with the gestures. Would it be possible to instead show the preview animation before the actual move?

Definitely, this will stay consistent with Loop’s existing settings :)
I just happened to have “Show preview when looping” turned off in the demo, which is why the windows moved instantly instead of showing the preview first.

> I was also wondering if custom gestures might be supported. For example, being able to assign a gesture to maximize or minimize windows. Personally, I’d love to disable Mission Control and Exposé so I could use a four-finger drag up/down gesture for maximize/minimize instead.

We’re definitely considering that as well. The rough idea I have right now is a dedicated Gestures tab in Settings, where you could assign a list of gestures to actions, similar to the Keybinds tab. For example, a two-finger slide (in any direction) to open the radial menu, a three-finger slide right to trigger a different action, or a four-finger swipe up to trigger another one.

I’m still very much in the brainstorming stage for how all of this would be configured, though, so that’s just the direction I’ve been thinking in so far. I’d love to hear any ideas you have on how that should work!

> And finally, will haptic feedback be part of this feature?

Absolutely :)

AlejandroVolkova:
@MrKai77 I think the implementation is brilliant. Invoking the radial menu from the title bar is a great choice, since it avoids false triggers in the app’s content view, which is one of the challenges with custom trackpad gestures.

Maybe it would feel even more natural to use a two‑finger pinch‑in to invoke the radial menu. I can imagine that once the radial menu is invoked via a trackpad gesture, it could stay persistent so the user can decide the next action, for example, a one‑finger swipe left to move a window to half the screen, similar to how the trigger key works currently with directional input from the trackpad.

On iPadOS there’s a gesture to minimize all other app windows when using “Windowed Apps.” Perhaps something similar could be achieved here by invoking the radial menu and swiping down with Three‑finger to trigger “Minimize others.”

Some actions might work best without invoking the radial menu at all. For instance, maximize or minimize could be triggered frictionlessly, just like swiping between spaces. That’s why I think Four‑Swipe‑Up → Maximize | Four‑Swipe‑Down → Minimize would be ideal, since those gestures can be triggered anywhere without interfering with app content. It would also be nice to keep the preview animation when maximizing, to stay consistent with Loop’s aesthetic and refinement.

Overall, the idea of a dedicated Gestures tab in Settings already sounds like the right direction. Other possible actions when invoked with the radial menu could be:

• Hide → Pinch out (Spread With Thumb And 3 Fingers)
• Center app → Three‑finger swipe up
• Stash → Three‑finger swipe right

It’s also worth mentioning that three‑finger gestures can sometimes cause conflicts when using the option Accessibility → Pointer Control → Trackpad Options… → Dragging style → Three Finger Drag 

Of course, all of these suggestions ultimately depend on how they feel in practice when testing the software :) They just came to mind from my past experiments with custom gestures, so thank you so much for being open to feedback. I really appreciate the chance to contribute ideas :) !

343max:
I just came looking here because I was wondering if I should implement this feature. Please please please make it so that I can use four finger gestures to trigger the actions. I'd love to see the implementation exactly as it is right now, but instead of doing a Modifier-Drag gesture I could just do a multitouch-drag without a modifier.

Loop.xcodeproj/project.pbxproj
```diff
@@ -310,8 +310,9 @@
 			packageReferences = (
 				2A28B6272EE5050C00A1E26B /* XCRemoteSwiftPackageReference "Defaults" */,
 				3ED0A7B72F21DF6800A58629 /* XCRemoteSwiftPackageReference "ZIPFoundation" */,
-				2AF9238C2F540B1300F467FD /* XCRemoteSwiftPackageReference "Scribe" */,
-				2AC4F62A2FDCAF7100CB2FBE /* XCRemoteSwiftPackageReference "Luminare" */,
+				2A847DFA2F5E49E90099E02A /* XCRemoteSwiftPackageReference "Scribe" */,
+				2A847DFD2F5E4A080099E02A /* XCRemoteSwiftPackageReference "Subsurface" */,
+				2AAC0B002FDD30370047C801 /* XCLocalSwiftPackageReference "../Luminare" */,
```
SenpaiHunters: Need to move Luminare back to the main branch or use a version that is not pinned locally.

Loop/Core/Multitouch/MultitouchTrigger.swift
```diff
@@ -0,0 +1,290 @@
...
+
+        radialMenuActionsObservationTask = Task { [weak self] in
+            for await _ in Defaults.updates(.enableRadialMenuCustomization, .radialMenuActions, initial: false) {
+                guard !Task.isCancelled, let self else { break }
+                radialMenuActions = RadialMenuAction.userConfiguredActions
+            }
+        }
+    }
+
+    func stop() {
```
SenpaiHunters, line 88: If Accessibility is revoked while gestures remain enabled in Defaults, `stop()` can leave conflicting macOS gestures disabled even though Loop is no longer handling gestures. `reconcileSystemGestures()` should consider actual runtime availability, not just the persisted `enableGestures` preference, or `stop()` should explicitly restore system gestures when the monitor is stopped.

Loop/Settings Window/Settings/Gestures/GestureItemView.swift
```diff
@@ -0,0 +1,249 @@
...
+                            IconView(action: action)
+
+                            Text(action.getName())
+                                .fontWeight(.regular)
+                                .lineLimit(1)
+                        } else {
+                            Image(systemName: "bolt.horizontal.fill")
+                                .foregroundStyle(.secondary)
+
+                            Text(String(localized: "No Action", comment: "Label shown for a gesture with no configured action"))
```
SenpaiHunters, line 181: `No Action` gestures are still treated as active gestures. They should not apply a window action, but they still register recognizers, can open Loop with `.noSelection`, start the gesture blocker, and count toward `SystemGestureManager` disabling conflicting macOS gestures. If `No Action` is meant to be an inert placeholder, these bindings should be filtered out before recognizer setup and system-gesture reconciliation; otherwise the UI should make clear that they still reserve/consume that gesture.

Loop/Core/Multitouch/MultitouchTrigger.swift
```diff
@@ -0,0 +1,290 @@
...
+        }
+    }
+
+    private func handleGestureEvent(_ event: SubsurfaceGestureEvent, fingerCount: Int) async {
+        switch event {
+        case let .swipe(swipe):
+            await handleSwipe(swipe, fingerCount: fingerCount)
+        case let .magnify(magnify):
+            await handleMagnify(magnify, fingerCount: fingerCount)
+        case .determining:
```
SenpaiHunters, line 155: Radial-menu gestures currently activate during Subsurface’s `.determining` phase, before Subsurface has recognized a swipe or magnify. Since `.determining` is emitted as soon as the configured finger count is present, resting the matching fingers on a titlebar can open Loop and start blocking native gesture events without crossing any gesture threshold. This contradicts the comment that Loop opens only after activation. If early preview is intentional, the behavior should be documented in UI/code comments; otherwise `activateGestureIfNeeded` should be gated until a resolved swipe/magnify `.began` event or an explicit displacement threshold.

SenpaiHunters (changes requested):
Also remember to update the README comparison table!

Loop/Settings Window/Settings/Gestures/GestureItemView.swift
mrkai77: `No Action` means the gesture is active but has no assigned action, so invoking it still opens the radial menu with a question mark. This matches the behavior of keybinds, where triggering a keybind linked to `No Action` still presents the radial menu with the question mark icon to show the user that it hasn’t been configured. Because of this, I think the existing UI already makes it clear that the gesture is reserved and consumed, and it should therefore participate in system-gesture reconciliation.

diverqent:
Hello!

My apologies for creating a duplicate request. I didn't find this thread when looking for it. Seems you've been on it for a while already.
Is this already available? I'm not really familiar with Github so I can't figure out whether it is or not but I would love to contribute as a tester.

Also: Am I reading correctly that the current implementation is to trigger it when hovering over the menubar, similar to Swish?
My request however was to be able to set custom swipe actions regardless of cursor screen position, **without** triggering the radial menu _and_ **without** showing previews. I'd like to be able to trigger the action instantly with a customised swipe action, such as "three finger swipe left" triggers "move current screen to left half of the screen" or "three finger swipe down" triggers a key shortcut combo or "three finger swipe down" triggers "minimize window", etc.
Will this be possible as well?

mrkai77:
Hey @diverqent! Yes, work on trackpad gestures has been ongoing for quite a while, and it’s already *mostly* done! There are still occasional crashes here and there, which is why it hasn’t been included in a public developer release yet.

Luckily, most of what you’re asking for is already supported! You can configure a gesture to open Loop’s radial menu, or assign it directly to a specific action, similar to assigning actions in the Keybinds tab. Each gesture can also be configured to work only over a window’s title bar or anywhere on screen, with a chosen number of fingers and gesture type (swipe, magnify, etc.).

The only part that won’t be included in the first iteration is choosing whether the radial menu or preview appears on a per-action basis. For now, gestures will follow your existing Loop-wide presentation settings.

As for when you’ll be able to start testing this, it’ll likely be available in developer builds in early to mid-September, as I’m quite busy with a few other things this month :)

diverqent:
That sounds amazing! Honestly, I can't wait to try it. I've been looking for an app that supports both a radial menu of some sorts for use with an external mouse, and track gestures for use with just the trackpad, both in one app. I was just playing around with the settings within Loop again and found out so much more than I did before. This app is powerful already, and I'm so excited because of it! 
Do you happen to have a Discord channel as well, were non-developers can also help to spot bugs or provide feedback? There isn't much I would personally change, but there are some little things here and there that I think could improve the experience overall and I would love to help. 

Either way, thank you for what you're doing! I really hope Loop will get the recognition it deserves.

PR #1065, opened 2026-03-09
⚡ Performance optimizations

MrKai77:
- Makes WindowRecords an `actor`, to prevent concurrent calls from crashing the app
- Made short-term static properties of `Window` cached inside a snapshot to prevent repeated calls that yield the same value (e.g. frame, isResizable)
- Updated codebase to work with the above changes accordingly (lots of `async`/`await` additions)
- Cleaned up some unused code


SenpaiHunters (approved):
I've had it running for the past three hours. I don't see any significant issues, such as major breaks or anything that restricts my workflow, nor have I noticed any sudden spikes while running it through instruments. I believe it's ready for a development release. You can either keep this branch and push a build from it or merge it with the main development branch. It's up to you, @MrKai77.

PR #1068, opened 2026-03-15
✨ Improved snapping on left, right, and bottom edges

### LINKED ISSUE #1066: ✨ Support Tiling windows by thirds vertically
### Feature Request Summary

I see there is a way to drag a window horizontally to fill up the space by 1/3 or 2/3. However, I don't see a way to do this vertically. I would like to achieve this for my vertical monitor.

For reference, Rectangle can do this today.

### Problem / Opportunity & Desired Solution

Problem:
- No 1/3 or 2/3 support for vertical layout
Solution:
- Add 1/3 or 2/3 support for vertical layout

### Alternatives Considered

Use Rectangle

### Visuals or References

_No response_

### Additional Context

_No response_

### Priority

Medium (important but not blocking)

### Final Checks

- [x] This is a single feature request (multiple feature requests should be reported individually).
- [x] I have looked to see if this is a duplicate of another feature request.
- [x] I can help with research or testing.
- [x] I can help code this feature.

jvanderen1:
## Description

Adds Rectangle-style vertical snapping to `WindowDirection+Snapping.swift`. When dragging a window to the left, right, or bottom screen edge, snap zones now include corners, halves, thirds, center thirds, and two-thirds cycling — providing full coverage of all tiling options via drag-to-snap.

Fixes #1066

### Changes

- Left/Right edges now have 7 vertical zones (by mouse Y position): corner quarters (~1%), top/bottom half (~1–6.3%), top/bottom vertical third (6.3–33%), and a center zone that defaults to `verticalCenterThird` with `topTwoThirds`/`bottomTwoThirds` cycling
- Bottom edge now has the same 7-zone layout applied horizontally: corner quarters, left/right half, left/right third, and horizontalCenterThird with leftTwoThirds/rightTwoThirds cycling
- Consolidated `processLeftSnap`, `processRightSnap`, and `processBottomSnap` into a single generic processEdgeSnap function with an `EdgeZoneDirections` configuration struct, eliminating code duplication
- Zone thresholds for corners (1/95) and halves (6/95) are derived from measurements of Rectangle's snapping behavior

## How has this been tested?

macOS (Sequoia 15.3.2)

Tested drag-to-snap on all three edges (left, right, bottom), verifying each of the 7 zones returns the correct `WindowDirection`:

- [x] Verified two-thirds cycling works when dragging from a third zone into the center zone
- [x] Left edge: corner quarters, `topHalf`, `topThird`, `verticalCenterThird`, `bottomThird`, `bottomHalf` zones
- [x] Right edge: corner quarters, `topHalf`, `topThird`, `verticalCenterThird`, `bottomThird`, `bottomHalf` zones
- [x] Bottom edge: corner quarters, `leftHalf`, `leftThird`, `horizontalCenterThird`, `rightThird`, `rightHalf` zones
- [x] Center zone two-thirds cycling (e.g. `topThird` → drag to center → `topTwoThirds`)
- [x] Top edge behavior unchanged (`topHalf` at edges, `maximize` in center)

<details><summary><h2>Video Demos</h2></summary>

https://github.com/user-attachments/assets/a782ba3a-a3c4-4b7e-99c1-3b3f2ecd1ab8

https://github.com/user-attachments/assets/8575982c-83f9-4ebe-8452-d9841ce95ba0

</details>

## Checklist

- [ ] I have made corresponding changes to the documentation if applicable

I used [Claude Opus 4.6 (Thinking)](https://www.anthropic.com/claude/opus) to help develop the snapping zone logic, zone thresholds, `EdgeZoneDirections` consolidation pattern, and code structure. All changes were reviewed and tested manually.

### DIFF
--- a/Loop/Window Management/Window Action/WindowDirection+Snapping.swift
+++ b/Loop/Window Management/Window Action/WindowDirection+Snapping.swift
@@ -8,6 +8,58 @@
 import Foundation
 
 extension WindowDirection {
+    private struct EdgeZoneDirections {
+        let nearCorner: WindowDirection   // ~0-1% from start edge
+        let half: WindowDirection         // ~1-6.3% from start edge
+        let third: WindowDirection        // ~6.3-33% from start edge
+        let centerDefault: WindowDirection // 33-67% center zone default
+        let farThird: WindowDirection     // ~6.3-33% from end edge
+        let farHalf: WindowDirection      // ~1-6.3% from end edge
+        let farCorner: WindowDirection    // ~0-1% from end edge
+        /// (third/twoThirds) pairs for cycling in the center zone
+        let cycleNear: (third: WindowDirection, twoThirds: WindowDirection)
+        let cycleFar: (third: WindowDirection, twoThirds: WindowDirection)
+
+        /// Left edge: Y axis, corners are top-left/bottom-left
+        static let leftEdge = EdgeZoneDirections(
+            nearCorner: .topLeftQuarter,
+            half: .topHalf,
+            third: .topThird,
+            centerDefault: .verticalCenterThird,
+            farThird: .bottomThird,
+            farHalf: .bottomHalf,
+            farCorner: .bottomLeftQuarter,
+            cycleNear: (third: .topThird, twoThirds: .topTwoThirds),
+            cycleFar: (third: .bottomThird, twoThirds: .bottomTwoThirds)
+        )
+
+        /// Right edge: Y axis, corners are top-right/bottom-right
+        static let rightEdge = EdgeZoneDirections(
+            nearCorner: .topRightQuarter,
+            half: .topHalf,
+            third: .topThird,
+            centerDefault: .verticalCenterThird,
+            farThird: .bottomThird,
+            farHalf: .bottomHalf,
+            farCorner: .bottomRightQuarter,
+            cycleNear: (third: .topThird, twoThirds: .topTwoThirds),
+            cycleFar: (third: .bottomThird, twoThirds: .bottomTwoThirds)
+        )
+
+        /// Bottom edge: X axis, corners are bottom-left/bottom-right
+        static let bottomEdge = EdgeZoneDirections(
+            nearCorner: .bottomLeftQuarter,
+            half: .leftHalf,
+            third: .leftThird,
+            centerDefault: .horizontalCenterThird,
+            farThird: .rightThird,
+            farHalf: .rightHalf,
+            farCorner: .bottomRightQuarter,
+            cycleNear: (third: .leftThird, twoThirds: .leftTwoThirds),
+            cycleFar: (third: .rightThird, twoThirds: .rightTwoThirds)
+        )
+    }
+
     static func getSnapDirection(
         mouseLocation: CGPoint,
         currentDirection: WindowDirection,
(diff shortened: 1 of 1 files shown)


Loop/Window Management/Window Action/WindowDirection+Snapping.swift
```diff
@@ -8,6 +8,58 @@
...
+        /// (third/twoThirds) pairs for cycling in the center zone
+        let cycleNear: (third: WindowDirection, twoThirds: WindowDirection)
+        let cycleFar: (third: WindowDirection, twoThirds: WindowDirection)
+
+        /// Left edge: Y axis, corners are top-left/bottom-left
+        static let leftEdge = EdgeZoneDirections(
+            nearCorner: .topLeftQuarter,
+            half: .topHalf,
+            third: .topThird,
+            centerDefault: .verticalCenterThird,
```
mrkai77: With this change (and below), `leftHalf`/`rightHalf` aren't reachable from their respective edges anymore, and `bottomHalf` isn't reachable from the bottom edge. Those are probably the most common actions for those edges, so I think they should be the center zone default instead of `verticalCenterThird`/`horizontalCenterThird`.

For making center thirds still accessible, I'm thinking that the corner and half zones could work as stepping stones that change what the larger zones return. So for example:

- Drag straight to center -> `half`
- Start from a corner, drag inward -> `twoThirds`
- Start from a half zone (or even third zone), drag inward -> `centerThird`

This way every action is still reachable, and the currentDirection cascades naturally as you drag through. Same pattern would apply to all three edges :)

Also interested in hearing if you have any other solutions! I haven't tested my idea above either; it could end up feeling unnatural, too.

MrKai77 (changes requested):

MrKai77:
Otherwise great work! This is totally a great start on adding more actions to window snapping! Just want to make sure we don't introduce any _significant_ breaking changes for existing users, such as the halves I mentioned above.

Loop/Window Management/Window Action/WindowDirection+Snapping.swift
jvanderen1: I went ahead and integrated your feedback into the PR. See the attached video snippet with the feedback. Can update the official PR description with the video snippet once approved.

https://github.com/user-attachments/assets/02292a43-9831-43fc-a328-bf9a6da821e1

I personally think this seems to work great! Slightly newer pattern than I am used to, but I think it covers the "edge" cases well (no pun intended). I would love to know if you have any proposed changes to the top edge. I think having `maximize` is desirable, but I do also think it's very different from the other edges. Not necessarily a bad / wrong thing, but could potentially be better.

Before submitting this PR, I can re-commit the changes underneath 1 commit. Just makes it easier to track changes for now.

Please let me know what you think.

jvanderen1:
@MrKai77 I appreciate the feedback provided! I agree that minimizing breaking changes would minimize risk for our current users. I don't know if this application has ever put proposed changes behind an experiment, but that may be helpful (or a separate configuration).

I think it would also be helpful for you or other users to try this out before submitting. I am able to build and run this myself, but if possible, I would like user feedback from folks actually trying it out.

Happy to take further feedback!

P.S. another side comment, I do notice that dragging against the top edge takes a *really* long time to compute compared to other edges. You can see my cursor struggling on the top edge in the new video snippets. Any known reason for this?

Loop/Window Management/Window Action/WindowDirection+Snapping.swift
jvanderen1: Looking closer at this, I inadvertently removed the 2 thirds trigger. I can't seem to snap into this state anymore. I will take a look during my spare time and see if I can resolve that.

mrkai77: No worries, thank you! I’ll review these changes as soon as I’m available as well :)

jvanderen1: I updated the implementation a bit and I gotta say, it feels really good to use!

https://github.com/user-attachments/assets/892dba96-2189-4174-b0fe-64711e912ebe

Loop/Window Management/Window Action/WindowDirection+Snapping.swift
```diff
@@ -8,59 +8,209 @@
...
+            return zones.cycleFar.twoThirds
+        }
+
+        // From a half/third outer zone → centerThird
+        let outerZones: [WindowDirection] = [
+            zones.half, zones.farHalf,
+            zones.third, zones.farThird
+        ]
+        if outerZones.contains(currentDirection) {
+            return zones.centerThird
```
mrkai77, line 214:
```suggestion
            if mousePos < centerMid - threshold {
                return zones.cycleNear.twoThirds
            } else if mousePos > centerMid + threshold {
                return zones.cycleFar.twoThirds
            }
            return zones.centerThird
```
This seems to fix a one-frame flicker when entering the center zone from a third/half zone. Without it, `centerThird` is always assigned first regardless of mouse position, then immediately overridden by the position check on the next event.

MrKai77 (changes requested):
This is working really well actually!

On the experiment point: we sometimes tuck more advanced features behind the Advanced tab or a defaults write flag to keep them out of reach for most users, but I don’t think that’s necessary here.

I’ve tested this locally and it feels very smooth. I’ll also tag @SenpaiHunters to give it a try if possible. Once merged, this will roll into Loop’s development build, where a few hundred beta testers can provide additional feedback before it reaches stable.

Regarding your side note: the issue is likely caused by this line:
https://github.com/MrKai77/Loop/blob/c2743567e2b8ddb88722f00c4703b73924c7834c/Loop/Core/WindowDragManager.swift#L111

You can also mitigate it by disabling “Suppress Mission Control” in Loop’s settings.

Overall, this is looking fantastic, really appreciate the work here :)

Loop/Window Management/Window Action/WindowDirection+Snapping.swift
jvanderen1: I went ahead and added the suggestion as a separate commit! If you want me to squash these commits or update and videos before merging, just let me know!

Thanks for all your support and enthusiasm working with me.

MrKai77 (approved):

PR #1087, opened 2026-04-25
✨ Add more `SkyLight` APIs for improved window/screen selection

### LINKED ISSUE #1056: 🐞 Stage Manager conflict: Loop resizes wrong window
### Bug Description

I have been facing a bug that occurs intermittently. Under the following conditions, Loop fails to resize the window under the cursor:

• Stage Manager: enabled
• Loop: Resize window under cursor: enabled
• Switching between apps using Tab
• Executing the trigger key to resize the window under the cursor

The issue observed is that, at times, Loop does not correctly detect the window currently in front. Instead, it seems to attempt resizing the application that was previously active.

What this causes is that, rather than resizing the current main window, Loop tries to resize the previously focused window. In the attached video, the failure can be observed, the preview border radius shown by Loop is significantly larger than that of normal windows. Following this, the applications swap positions because Loop does not execute the resize action on the window currently under the cursor.

This behavior can be sort of simulated when executing the trigger key while the cursor is outside of a focused window (for example, on the desktop).
It is worth noting that this bug only occurs when using Stage Manager.

### Affected Scope

User interface

### Steps to Reproduce

1. Enable Stage Manager.
2. Enable Loop: Resize window under cursor.
3. Switch between applications using Tab.
4. Execute the trigger key to resize the window under the cursor.
5. Observe the behavior when the active window changes.

### Reproducibility

Sometimes (30–70%)

• Expected:
Loop should detect and resize the current front window under the cursor.
• Actual:
Loop intermittently resizes the previously focused window instead, showing an oversized preview border and causing apps to swap positions.

https://github.com/user-attachments/assets/4dd9b029-c6e8-43e5-a8e2-6d1c56c960ba

### Severity

Major (workaround exists)

### macOS Version

Tahoe 26.4 (beta)

### Loop Version

Version 🧪 1.4.3 (1709)

Yes

### Additional Context

_No response_

### Final Checks

- [x] I can help with further investigation.

MrKai77:
Adds two private SkyLight APIs: `SLSCopyBestManagedDisplayForPoint` and `SLSFindWindowByGeometry`.

`SLSCopyBestManagedDisplayForPoint`
Picks the display at a given point using the system's own selection logic instead of our manual bounds iteration. Not really a performance win, but the value is that we match WindowServer's tie-breaking at boundaries, so we can't drift from it (in practice the two seem to agree, though matching the source-of-truth is cheap insurance). Falls back to the old iteration if the symbol fails to load.

`SLSFindWindowByGeometry`
Finds the window at the mouse position. Previously we'd ask AX for the UI element under the cursor, or fall back to walking `CGWindowListCopyWindowInfo` and taking the first frame that contained the point. The fallback was unreliable, as it _could_ skip the frontmost window and return one obscured beneath the actual target. SkyLight's geometry hit-test gives us the right window directly. If this symbol is unavailable, we still fall back to the old system.

### DIFF
--- a/Loop/Extensions/NSScreen+Extensions.swift
+++ b/Loop/Extensions/NSScreen+Extensions.swift
@@ -29,6 +29,13 @@ extension NSScreen {
         let mouseLocation = NSEvent.mouseLocation
         let screens = NSScreen.screens
 
+        // Ask SkyLight first so we match WindowServer's tie-breaking at display boundaries.
+        if let primary = screens.first,
+           let displayID = SkyLightToolBelt.bestManagedDisplayID(forCGPoint: mouseLocation.flipY(screen: primary)),
+           let match = screens.first(where: { $0.displayID == displayID }) {
+            return match
+        }
+
         // CGRect.contains uses half-open intervals [minX, maxX) × [minY, maxY),
         // excluding points on the maxX/maxY edges. So we cannot use frame.contains(_:)
         return screens.first {
--- a/Loop/Private APIs/SkyLightSymbolLoader.swift
+++ b/Loop/Private APIs/SkyLightSymbolLoader.swift
@@ -81,6 +81,15 @@ extension SkyLightSymbolLoader {
     @available(macOS 26.0, *)
     static let SLSWindowIteratorGetResolvedCornerRadii: SLSWindowIteratorGetResolvedCornerRadiiFunc? = loadSymbol("SLSWindowIteratorGetResolvedCornerRadii")
 
+
+    /// Returns the UUID (`CFString`) of the display containing `point`, using the same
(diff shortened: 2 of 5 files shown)


SenpaiHunters (approved):
We're missing any error checking for `SLSFindWindowByGeometry` as you're doing

```swift
_ = SLSFindWindowByGeometry(cid, 0, 1, 0, &screenPoint, &windowPoint, &hitWindowID, &windowCID)
return hitWindowID != 0 ? hitWindowID : nil
```

If we ignore errors, that's acceptable. I'm uncertain about what it would return; if it yields nothing useful, we can disregard it. Additionally, I see no changes or updates needed. If we don't need to log the SLS failure, it's ready to commit.

PR #1088, opened 2026-04-28
✨ Add `SLSGetWindowLevel` inside window validation checks

MrKai77:
Previously, the window level check only existed in `Window.fromWindowInfo`, since that was the only public way to retrieve a window's level. With SkyLight, we can now fetch a window's `CGWindowLevel` directly via `SLSGetWindowLevel`, allowing the check to live in `Window`'s main initializer and apply to every code path. This lets Loop automatically reject intentionally floating windows (such as the notch window from Alcove and likely other notch apps) instead of relying on per-app blacklisting to keep Loop from manipulating them.

Note that the original `Window.fromWindowInfo` check has been kept as an early bail-out, as it lets Loop reject invalid windows up front using data already in hand, avoiding an unnecessary `SkyLight` call :)

### DIFF
--- a/Loop/Private APIs/SkyLightSymbolLoader.swift
+++ b/Loop/Private APIs/SkyLightSymbolLoader.swift
@@ -104,6 +104,9 @@ extension SkyLightSymbolLoader {
     typealias SLSSetWindowBackgroundBlurRadiusFunc = @convention(c) (_ connection: SLSConnectionID, _ wid: CGWindowID, _ radius: Int) -> OSStatus
     static let SLSSetWindowBackgroundBlurRadius: SLSSetWindowBackgroundBlurRadiusFunc? = loadSymbol("SLSSetWindowBackgroundBlurRadius")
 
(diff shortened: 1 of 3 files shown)


SenpaiHunters (approved):

AlejandroVolkova:
@MrKai77 This new implementation is awesome :) ! I’m almost 100% sure it also fixes issue #1056 I reported earlier. It would be great if you guys could verify this with the current build, since I can’t trigger it anymore on my device.

Loop: 1.4.3 (1723)
macOS 26.4.1

MrKai77:
Thanks for the update @AlejandroVolkova! I can’t reproduce it anymore either, but it’s likely that it was instead PR #1087 fixed it. It introduced a new SkyLight check called `SLSFindWindowByGeometry`. Essentially, Loop can now _directly_ ask the system "which window is located here?" instead of relying on accessibility APIs or, worse, iterating through every window’s reported position and guessing which one is actually under the cursor. I’ll go ahead and close the issue, great to hear it’s resolved :)

PR #1156, opened 2026-09-20
Add a Next/previous screen order setting: Z-shaped, clockwise or counterclockwise

### LINKED ISSUE #1000: ✨ Option to make Next/Previous display movement follow clockwise order
### Bug Description

These are my display arrangements and the order I would expect when moving a window to the next / previous display in a clockwise direction, as illustrated below:

Image

In this layout, I expect the window to move between displays following the clockwise order shown by the numbered arrows.

Actual behavior

However, this is what actually happens:

Image

The window traversal order does not follow the physical clockwise arrangement of the displays and instead jumps between displays in an unintuitive order.

### Affected Scope

User interface

### Steps to Reproduce

1. I think you may need at least 4 displays for test.
2. Arrange displays in two lines.
3. Move window to next display/previous display.

### Reproducibility

Always

Expected behavior

When moving a window to the next or previous display, the window should follow the physical display layout in a clockwise order.

Actual behavior

When moving a window to the next or previous display, the window moves between displays in a non-clockwise and unintuitive order.

_No response_

### Severity

Critical (major functionality broken)

### macOS Version

macOS 15.7.3

### Loop Version

Version 🧪 1.4.1 (1673)

Yes

### Additional Context

_No response_

### Final Checks

- [ ] I can help with further investigation.

anandghegde:
## Description

Adds an advanced setting for the order `Next/Previous Screen` cycles through screens: **Z-shaped**, **Clockwise** or **Counterclockwise**.

This implements what you described in #1000:

> I'm thinking that it might make more sense to expose an advanced option that lets them choose how "Next/Previous Screen" determines the next screen: using a Z-shaped order, clockwise, or counter-clockwise.

and, for the 3×3 case raised in the same thread:

> I'd expect it would cycle through the outer displays clockwise first, then move on to the inner displays, again in a clockwise order :)

Closes #1000

### What changed

- `ScreenOrder` (new, `Loop/Utilities/ScreenOrder.swift`) — the three orders, as a `Defaults.Serializable` enum. Its one entry point is `sorted(_:frame:)`, which takes anything plus a way to get its frame, so the ordering has no dependency on `NSScreen`.
- `ScreenUtility.getOrderedScreens()` now just asks `Defaults[.screenOrder]` to sort `NSScreen.screens`. The existing "vertically stacked Z" comparator moved into `ScreenOrder.zShaped` unchanged, so **`.zShaped` is byte-for-byte the current behaviour**, and it is the default — nothing changes for anyone who doesn't touch the setting.
- `Defaults.Keys.screenOrder`, in the `// Advanced` block, `iCloud: true` like its neighbours.
- A `Screens` section in `AdvancedConfigurationView`, between General and Radial Menu, holding a single `LuminarePickerMenu`.

### How the rotational orders work

Ordering purely by angle around the centre of the layout (the approach suggested in the issue thread) breaks on the two arrangements that matter most: a single row, where every screen sits at ~0° or ~180° and the middle ones land arbitrarily, and 3×3, where the centre screen has no defined angle at all.

So instead the screens are peeled off in rings:

1. Take the screens on the outside of the arrangement — the convex hull of their centres, keeping centres that lie *on* a hull edge so a straight row stays one ring instead of losing its middle screens.
2. Walk that ring by angle about its own centroid, starting from the screen nearest the top left.
3. Repeat with whatever is left inside.

A row or column has no inside, so it is a single ring and gets ordered along its dominant axis instead of by angle — left-to-right for a row, top-to-bottom for a column, since a row is the top of the circle and a column is its right-hand side. Two screens are a single cycle either way round. 3×3 gives the eight outer screens clockwise from the top left, then the middle one.

All of this runs on `frame`, i.e. AppKit's global coordinates where **y points up**, which is why ascending `atan2` is counterclockwise here.

## How has this been tested?

macOS 26.4.1, Xcode 26.4.1, branched off `develop` @ df26d56.

- [x] `LoopTests/ScreenOrderTests.swift` — 14 tests, all passing (`xcodebuild test -only-testing:LoopTests/ScreenOrderTests -skipMacroValidation -parallel-testing-enabled NO`). They cover a row, a row of mismatched-size displays whose centres don't line up, a column, a 2×2 block in both directions, the 3×3 from the issue, one screen, two screens, and two invariants that matter for this feature specifically: no screen is ever dropped or duplicated, and starting anywhere and stepping forward reaches every screen before wrapping.
- [x] Built and ran the app, confirmed the new section renders in Advanced, that all three options appear, and that choosing one persists (`defaults read com.MrKai77.Loop screenOrder` → `1` after picking Clockwise).
- [x] `swiftformat --lint` clean on all five files.

**Please read this bit:** I have one display on this machine, so I have not watched a window actually hop around a real multi-display setup. That is the reason the ordering is a pure function of frames rather than something reaching into `NSScreen` — every arrangement above is exercised as a unit test on synthetic frames, and `ScreenUtility` only supplies `\.frame`. But a test is not a pair of eyes, and I'd rather say so than imply coverage I don't have. @Caffeine19, if you still have the two-row layout from the issue, this is the branch to try.

One deliberate omission: I did not hand-edit `Loop/Localizable.xcstrings`. Every entry in it is fully translated Crowdin output with no English-only rows, so adding partial entries by hand seemed more likely to disrupt the sync than help. The four new strings ("Screens", "Next/previous screen order", and the three case names) are all in code and marked for extraction. Happy to add them if you'd prefer.

<details><summary><h2>Screenshots</h2></summary>

The new Screens section in Advanced, with the Next/previous screen order menu open showing Z-shaped, Clockwise and Counterclockwise

</details>

## Checklist:

- [x] I have made corresponding changes to the documentation if applicable

Claude (via Claude Code) was used, and used heavily: it wrote the first draft of `ScreenOrder.swift`, the test file, and this description.

What I did with that output, since the policy asks:

- I picked the ring-peeling approach myself after the model's first attempt did plain angular sorting, which I rejected because it mishandles a single row and has nothing sensible to say about the centre screen in 3×3 — the exact case you called out in the issue.
- I hand-traced the expected order for every test case rather than trusting the generated expectations. One of them was wrong: `aRowOfDifferentlySizedScreensIsStillWalkedLeftToRight` was generated expecting `["main", "side", "laptop"]`, but the rotate-to-top-left step actually produces `["laptop", "main", "side"]`. The code was right and the test was wrong, so I fixed the test.
- While getting the tests to run I hit `The test runner hung before establishing connection`. Rather than assume it was my change, I ran an existing suite (`KeybindResolverTests`) and saw the same failure, which pointed at xcodebuild's parallel test clones tripping Loop's single-instance terminate broadcast — hence `-parallel-testing-enabled NO` above. Worth knowing independently of this PR.
- I built the app, clicked through the new setting, and checked the written preference myself.

I understand every line here and am happy to explain or change any of it.


mrkai77:
Thanks @anandghegde! I can’t get around to testing this at the moment, but I actually started working on something similar a few months ago on the `kai/clockwise-screen` branch. It wasn’t working perfectly for me, which is why I never merged it, and yours will likely work better!

One thing I do think would be worth adapting from my branch is the [UI](https://github.com/mrkai77/Loop/commit/9d54e1e8aef7b61d9740f13fbdb87af3fb338967#diff-dec3a983fe9ec40b5cf66d4682e4c64785b7f010e34789165ec63cebc78983a3R149). It uses a `LuminarePicker` with custom SF Symbols to help illustrate what each setting would look like.

Would you be able to update the UI to use something similar, or transfer the relevant code over from that branch?

